PR #1, opened 2018-06-08
Initial spec import

gkellogg:
Spec updated from CG for WG.


iherman:
I hope I understand your question correctly...

There is a preview tool but it _works for PR-s only_. 
The tool must be installed called "PR Preview", developed by Tobie Langel. You also need a config file in the repository called ".pr-preview.json", which, for this repo, should contain:

{
    "src_file": "index.html",
    "type": "respec"
}

The tool is hugely useful. What it does, is that it runs a respec via (I believe) node.js with the result somewhere on the AWS cloud and changes the PR's comment to add a reference to the generated version and the diff file (see, eg, [a Publ WG repo](https://github.com/w3c/wpub/pull/214) as an example). It may make mistakes with diagrams that, for some reasons, it does not always find, but that is a minor problem compared to the benefits it brings.

***But***: It is already installed because, by default, it is installed on all W3C repositories! And I have already added the extra config files 😄

iherman:
Hm. This should work, because the tool to preview is automatically installed for all W3C repos, including this one (and I have added the necessary config files). I _suspect_ that it has problems with respec; afaik, if the respec process does not complete for whatever reasons, then there is no result appearing here.

Can you check it off separately? A way to do that is to run something like below, with the obvious changes of the parameters:

```
https://labs.w3.org/spec-generator/?type=respec&url=https://w3c.github.io/wpub/?specStatus=FPWD;publishDate=2018-01-02;includePermalinks=false
```

Cc: @tobie

tobie:
Hey folks, unfortunately, I don't have the time to do any support for pr-preview these days. I am considering offering support as a paid service however, so let me know if this is something you might be interested in.

iherman:
@gkellogg,

I believe the issues is with

```
<script type="text/javascript" src="../json-ld-wg/common/common.js" class="remove"></script>
```

I am afraid the 'common' subdirectory will have to be duplicated for each document, and it cannot be kept at one place. B.t.w., the lab generator cannot go to that directory and, actually, this would also create, at some point, a problem with Echidna.

:-(

iherman:
@gkellogg I have made a version and a separate PR #2 where:

- I have copied the content of the 'common' directory to the repository
- Changed the references in `index.html` (including the two html import statements)
- While I was at it, I also updated the Patent Policy reference in the respec header, as well as added a reference to the 2018 process.

The PR preview works now in #2. I propose to close this one, and merge #2.

davidlehn:
Do other specs have similar issues and have a good solution to keep common files in sync? Maybe just a README file in each spec common dir noting that people should edit the main wg common files and sync changes to each spec? Or maybe it's just not a big deal to do by hand?

gkellogg:
It sort of misses the point of having common files, if they need to be duplicated, which is unfortunate. Certainly, we'll need to add a README that changes to any file in /common must be replicated to other repositories; as Ivan noted, they don't change too often. We should probably keep the main ones in json-ld-wg. Perhaps we can figure out a Travis CI check to make sure they haven't diverged.

I'll close PR #1 and replicate these changes to json-ld-api and json-ld framing PRs.

PR #35, opened 2018-07-10
Use IRIs instead of semantically-named blank node identifiers

### LINKED ISSUE #27: Graph id-s in the examples
In all examples in section 4.15 on Named Graphs, the id-s of named graphs are blank nodes. This is not at all the way named graphs are used in practice and may be misleading for the user. We should use real URL-s whenever possible.

As a misleading example consider 4.15.1, where the blank node ID `_:claim` is used. For a reader who is not fully versed in RDF this _may_ give the impression that the blank node's ID has some semantic meaning, just like URL-s may have. A similar misleading example is the usage of `_:manu` and `_:gregg` in example 79.

gkellogg:
in named graph examples.

Fixes #27.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/35.html" title="Last updated on Aug 5, 2018, 7:42 PM GMT (3470dfe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/35/751cd60...3470dfe.html" title="Last updated on Aug 5, 2018, 7:42 PM GMT (3470dfe)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2538,7 +2538,7 @@ specified. The full <a>IRI</a> for
       "name": "http://xmlns.com/foaf/0.1/name",
       "knows": "http://xmlns.com/foaf/0.1/knows"
     },
-    ****"@id": "_:graph",
+    ****"@id": "http://example.org/foaf-graph",
     "generatedAt": "2012-04-09",
     "@graph":**** [
       {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2743,7 +2744,7 @@ <h3>Graph Containers</h3>
     </pre>
 
     <p>The example above expresses a <a>named graph</a> that is identified
-      by the <a>blank node identifier</a> <code>_:claim</code>. That
+      by the <a>blank node identifier</a> <code>_:b0</code>. That
```
dlongley: Maybe we should have another example before this one with no blank node identifier. Then, this example can follow and say that the blank node identifier only exists so the graph can be referenced within the same document and that the blank node identifier is meaningless otherwise. I think that would help people understand it.

Issue #34, opened 2018-07-08
@type as @container:@set?


gkellogg:
The purpose of the `@container:@set` functionality (AFAIU) is to ensure that the output is consistent in shape. Thus if there can ever be multiple values, the structure is always an array.

There are two situations in which this functionality could be desirable but is currently not possible:

1.  `@type` As it's a keyword, we can only alias it (e.g. as `type`) but not define it to have `@container:@set` functionality. Meaning that there's a gotcha waiting to happen for ontologies that require or use multiple classes for a single resource instance. See [playground](https://json-ld.org/playground/#startTab=tab-compacted&json-ld=%7B%22%40context%22%3A%7B%22cls%22%3A%7B%22%40id%22%3A%22http%3A%2F%2Fexample.org%2Fclass%22%7D%2C%22type%22%3A%7B%22%40id%22%3A%22%40type%22%2C%22%40container%22%3A%22%40set%22%7D%7D%2C%22%40id%22%3A%22foo%22%2C%22type%22%3A%22cls%22%7D&context=%7B%22%40context%22%3A%7B%22cls%22%3A%7B%22%40id%22%3A%22http%3A%2F%2Fexample.org%2Fclass%22%7D%2C%22type%22%3A%7B%22%40id%22%3A%22%40type%22%2C%22%40container%22%3A%22%40set%22%7D%7D%7D)
2.  `@context` Less useful, but `@context` will also compact to a single string/object when there might be multiple contexts.  See [playground](https://json-ld.org/playground/#startTab=tab-compacted&copyContext=true&json-ld=%7B%22%40context%22%3A%5B%7B%22cls%22%3A%7B%22%40id%22%3A%22http%3A%2F%2Fexample.org%2Fclass%22%7D%2C%22type%22%3A%7B%22%40id%22%3A%22%40type%22%2C%22%40container%22%3A%22%40set%22%7D%7D%5D%2C%22%40id%22%3A%22foo%22%2C%22type%22%3A%22cls%22%7D)

`@context` modifying itself seems particularly strange, but the inconsistency for `@type` seems solvable if the restrictions in its definition were loosened?

Original issue [@type as @container:@set? #512](https://github.com/json-ld/json-ld.org/issues/512).

(Edit: Note that the playground links demonstrate that these do not work as one might expect, due to the keyword definitions of @type and @context)

dlongley:
Seems like a useful feature to allow `@container: @set` for `@type`.

[...]

Issue #26, opened 2018-07-01
Explanation of examples


iherman:
The various examples are shown in other syntaxes in a somewhat inconsistent manner: most of them use the result of the expansion but some of them use a table figuring, essentially, the triples.

Also, the usage of the expansion algorithm to illustrate the example is not really helpful, mainly at first reading. Most of the examples are in sections 3 and 4, whereas the expansion algorithm is defined at the very end of section 4 (in 4.25). Also, the result of the expansion algorithm, while o.k. for machine processing, is not always intuitive: e.g., everything is put into arrays, even in cases when this is fairly unexpected for a JSON reader.

I would propose two editorial changes in this respect.

1. Move sections 4.25-4.27 into a separate top level sections. These are important for the handling of JSON-LD files, but not necessary important for the understanding of JSON-LD.
2. For _every_ examples, there should be an equivalent of the example in the expanded form, in a table with the triples, in turtle (as close to the JSON-LD structure as possible) and, possibly, as graphs. Not all of them would appear on the screen at the same time but, rather, the reader could choose what to see with some tabs.  

(A good example for such an approach is in the [OWL Primer](https://www.w3.org/TR/owl2-primer/), where there is tab at the top that governs what syntax the reader would like to see.)

iherman:
Explicit examples for framing or flattening are of course a separate issue.

I think that consistency is important: _all_ examples (excep for the flattening/expansion ones, of course) _must_ have the same series of examples, whether it makes a lot of sense or not. And relying on playground is not an option for me: I may not always be on-line (I reviewed the document, and gathered these comments, on a plane) and playground may be down...

If we can have a control over each example easily, I would also be in favour of that. And if the document is large: let us be it. We will sill be smaller than the HTML doc, and infinitely more readable:-)

iherman:
B .t.w., on my last remark: if the document is too large as one HTML file, we can consider cutting the text into several files (within on recommendation) like HTML or SVG does it...

BigBlueHat:
@iherman do you know of any specs that use multiple representations/displays at once--thinking something like this would be nice https://schema.org/Person#Invoice-gen-43

gkellogg:
[...]

Issue #7, opened 2018-06-30
support @values for describing multidimensional containers


gkellogg:
Allows a term definition to include an `@values` block to describe structured values, such as for GeoJSON

  ### Example

  ``` json
  {
    "@context": {
       "coordinates": {
          "@id": "geojson:coordinates",
          "@container" : "@list",
          "@values" : { 
             "@type" : "geojson:Coordinate",
             "@container" : "@set",
             "@values" : [
                 {"@type" : "xsd:double", "@id":"geo:longitude"},
                 {"@type" : "xsd:double", "@id":"geo:latitude"}
             ]
          }
       }
    },
    "@graph" : [{
     "@id" : "ex:LineString1",
      "coordinates" : [
            [
              3.1057405471801753,
              51.064216229943476
            ],
            [
              3.1056976318359375,
              51.063434090307574
            ]
      ]
    }]
  }
  ```

  Would transform to (and vice versa):

  ``` turtle
  ex:LineString1 geojson:coordinates _:b0 .
  _:b0 rdf:first _:b1 .
  _:b1 a geojson:Coordinate ;
        geo:longitude "3.105740547180175E0"^^xsd:double ;
        geo:latitude "5.106421622994348E1"^^xsd:double .
  _:b0 rdf:rest _:b2 .
  _:b2 rdf:first a geojson:Coordinate ;
        geo:longitude "3.1056976318359375"^^xsd:double ;
        geo:latitude "51.063434090307574"^^xsd:double .
  _:b2 rdf:rest rdf:nil .
  ```

Original issue is [JSON-LD 1.1 Feature Request: support @values for describing multidimensional containers (list of lists) #397](https://github.com/json-ld/json-ld.org/issues/397)

pietercolpaert:
Appologies -- Wrongly assumed opening this issue meant it was going to be adopted in JSON-LD 1.1 - https://twitter.com/pietercolpaert/status/1013100062879178754

Would love this proposal to be taken into account for the 1.1 spec though!

p1d1d1:
Any thoughts how this could map to GeoSPARQL?

Fak3:
refs https://github.com/json-ld/json-ld.org/issues/52

azaroth42:
+1 to the issue, we should discuss the proposed solution.

azaroth42:
[...]

PR #41, opened 2018-08-05
Describe lists containing lists, as motivated by GeoJSON.

### LINKED ISSUE #36: Support recursive @list usage
From [Should we support lists of lists?](https://github.com/json-ld/json-ld.org/issues/52)

We did not support recursive RDF lists in 1.0 due to perceived complexity and lack of a use case. Since then there are numerous use cases that can make use of recursive lists in JSON-LD, and it goes further towards full idiomatic support of RDF 1.1 in JSON-LD.

Changes to the syntax document are minimal, there are substantial API implications.

gkellogg:
Fixes #36.

This is implemented in w3c/json-ld-api#16.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/41.html" title="Last updated on Aug 11, 2018, 9:48 PM GMT (b754d86)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/41/751cd60...b754d86.html" title="Last updated on Aug 11, 2018, 9:48 PM GMT (b754d86)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2299,7 +2299,11 @@ specified. The full <a>IRI</a> for
 -->
 </pre>
 
-<p>The implementation of <a>lists</a> in RDF depends on linking anonymous nodes together using the properties <code>rdf:first</code> and <code>rdf:rest</code>, with the end of the list defined as the resource <code>rdf:nil</code>. This can be represented as triples, as the following example shows, but the results are somewhat unwieldy.</p>
+<p>The implementation of <a>lists</a> in RDF depends on linking anonymous nodes
+  together using the properties <code>rdf:first</code> and
+  <code>rdf:rest</code>, with the end of the list defined as the resource
+  <code>rdf:nil</code>. This can be represented as triples, as the following
+  example shows:</p>
 
 <table class="example">
 <thead><tr>
@@ -2346,7 +2350,7 @@ specified. The full <a>IRI</a> for
 </tbody>
 </table>
 
-<p>Consequently, most RDF serializations (including JSON-LD) provide a syntactic shortcut for these lists. In Turtle, the graph would be expressed as follows:</p>
+<p>JSON-LD provides a syntactic shortcut for these lists. In Turtle, the graph would be expressed as follows:</p>
 
 <pre class="example" data-transform="updateExample"
      data-content-type="text/turtle"
@@ -2358,7 +2362,11 @@ specified. The full <a>IRI</a> for
 -->
 </pre>
 
-<p class="changed">In JSON-LD 1.1, lists of lists, where the value of a <a>list object</a>, may itself be a <a>list object</a> recusively, are fully supported. For example, in <em>GeoJSON</em> (see [[RFC7946]]), <em>coordinates</em> are an ordered list of <em>positions</em>, which are represented as an array of two or more numbers. For example:</p>
+<p class="changed">In JSON-LD 1.1, lists of lists, where the value of
+  a <a>list object</a>, may itself be a <a>list object</a> recusively, are
+  fully supported. For example, in <em>GeoJSON</em> (see [[RFC7946]]),
+  <em>coordinates</em> are an ordered list of <em>positions</em>, which are
+  represented as an array of two or more numbers. For example:</p>
 
 <pre class="example changed"
      data-content-type="application/json"
@@ -2381,7 +2389,10 @@ specified. The full <a>IRI</a> for
 }
 </pre>
 
-<p class="changed">For this example, it's important that the values expressed within coordinates maintain their order, which requires the use of embedded list structures. In JSON-LD 1.1, we can express this using recursive lists, by simply adding the appropriate context definion:</p>
+<p class="changed">For this example, it's important that the values
+  expressed within coordinates maintain their order, which requires the use of
+  embedded list structures. In JSON-LD 1.1, we can express this using recursive
+  lists, by simply adding the appropriate context definion:</p>
 
 <pre class="example changed"
      data-content-type="application/json"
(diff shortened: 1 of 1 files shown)


iherman:
@gkellogg,

I do not have a real issue with the changes, so we can merge it; my comments are purely of an editorial nature and reflect a matter of taste... Take them as you wish, salt 'n pepper them at taste...

- I would separate the parts on sets and lists in two subsections. The two notions have a different feel and setting, and it would increase readability
- The session after the table reprsenting lists in RDF you have a section that begins with "Consequently...", which is a bit out of place in this document, we usually do not analyze other serializations. I would just give the example in Turtle, and leave that (and, also, the "unwieldy" comment before the table) out. In any case, as I already raised in #26, I would think we will have to rethink how we represent _all_ our examples anyway.
- You say "In JSON-LD 1.1, we can express this using recursive lists, by simply adding the appropriate context definion". This may be understood as this being the _only_ way of doing so, which is not true. One can use repeated `@list` keywords for each series of coordinates, it is just unreadable...

gkellogg:
> I would separate the parts on sets and lists in two subsections. The two notions have a different feel and setting, and it would increase readability

They are both about arrays and ordering, which is why they were together. We do need to do a comprehensive reorganization, but I'd rather not do that on a piecemeal basis.

> The session after the table reprsenting lists in RDF you have a section that begins with "Consequently...", which is a bit out of place in this document, we usually do not analyze other serializations. I would just give the example in Turtle, and leave that (and, also, the "unwieldy" comment before the table) out.

Noted.

> In any case, as I already raised in #26, I would think we will have to rethink how we represent all our examples anyway.

Indeed, but this is a somewhat daunting task, for which I could use some help. In particular, we probably need to use something like [jsonld-vis](https://github.com/scienceai/jsonld-vis) to simplify creating such graphs, although this may prove to be a problem for non-HTML use.

iherman:
Hey @gkellogg 

> They are both about arrays and ordering, which is why they were together. We do need to do a comprehensive reorganization, but I'd rather not do that on a piecemeal basis.

I was not clear, sorry about that. I meant deeper subsections within this subsection. But I agree that this can wait, subject to a reorganization

> Indeed, but this is a somewhat daunting task, for which I could use some help. In particular, we probably need to use something like jsonld-vis to simplify creating such graphs, although this may prove to be a problem for non-HTML use.

It is complicated indeed. The table/expanded/TriG/n-quads versions may not be that bad just lot of writing; if the graphs prove to be too much work, we can drop that one.

(B.t.w., for some reasons, example 57 does not appear in the preview. I presume it will be there for the merged version.)

azaroth42:
+1 to reorganization, and +1 to not doing it as part of this PR :)
Also ... +1 to the PR.

gkellogg:
I'd like to handle this PR and w3c/json-ld-api#16, which ideally could use an implementation, but should at least have a review.

simonstey:
I already pinged Axel and I'll give it a read tmrw!

index.html
```diff
@@ -2301,9 +2299,187 @@ <h2>Sets and Lists</h2>
...
+     title="An ordered collection of values in Turtle">
+<!--
+@prefix foaf: <http://xmlns.com/foaf/0.1/> .
+
+<http://example.org/people#joebob> foaf:nick ("joe" "bob" "jaybe") .
+-->
+</pre>
+
+<p class="changed">In JSON-LD 1.1, lists of lists, where the value of
+  a <a>list object</a>, may itself be a <a>list object</a> recusively, are
```
simonstey: s/recusively// (I don't think we need to say that)

index.html
```diff
@@ -2301,9 +2299,187 @@ <h2>Sets and Lists</h2>
...
+  fully supported. For example, in <em>GeoJSON</em> (see [[RFC7946]]),
+  <em>coordinates</em> are an ordered list of <em>positions</em>, which are
+  represented as an array of two or more numbers. For example:</p>
```
simonstey: s/numbers. For example:/numbers:/

BigBlueHat:
@sgillies as the primary author of [GeoJSON-LD](https://github.com/geojson/geojson-ld) your thoughts on this would be most appreciated!

index.html
```diff
@@ -2301,9 +2299,187 @@ <h2>Sets and Lists</h2>
...
+  "type": "Feature",
+  "bbox": [-10.0, -10.0, 10.0, 10.0],
+  "geometry": {
+    "type": "Polygon",
+    "coordinates": [
+        [
+            [-10.0, -10.0],
+            [10.0, -10.0],
+            [10.0, 10.0],
+            [-10.0, -10.0]
```
sgillies: I feel that GeoJSON examples can be made less confusing with respect to longitude, latitude ordering by using real-world coordinates like `[-105.615115, 40.255014]` (latitude: 40.255 degrees N, longitude: 105.615 degrees W).

index.html
```diff
@@ -2301,9 +2299,187 @@ <h2>Sets and Lists</h2>
...
+<pre class="example changed"
+     data-content-type="application/json"
+     data-transform="updateExample"
+     title="Coordinates expressed in JSON-LD">
+{
+  "@context": {
+    "@vocab": "https://purl.org/geojson/vocab#",
+    "type": "@type",
+    "bbox": {"@container": "@list"},
+    "coordinates": {"@container": "@list"}
```
sgillies: In https://github.com/geojson/geojson-ld/issues/34 we discussed an `@ndarray`, but `@list` looks just as good, or better in that it's easier to explain and understand.

index.html
```diff
@@ -2301,9 +2299,187 @@ <h2>Sets and Lists</h2>
...
+  }],
+  "https://purl.org/geojson/vocab#geometry": [{
+    "@type": ["https://purl.org/geojson/vocab#Polygon"],
+    "https://purl.org/geojson/vocab#coordinates": [{
+      "@list": [{
+        "@list": [
+          {"@list": [{"@value": -10.0}, {"@value": -10.0}]},
+          {"@list": [{"@value": 10.0}, {"@value": -10.0}]},
+          {"@list": [{"@value": 10.0}, {"@value": 10.0}]},
+          {"@list": [{"@value": -10.0}, {"@value": -10.0}]}
```
sgillies: This is fine. Tagential question: is there a way to indicate to JSON-LD users that the first element of these 2-tuples is longitude and the second is latitude?

gkellogg: Not really. If the vocabulary were modeled in OWL, there may be some way to do this at that level, @iherman may know better.

index.html
gkellogg: Note that the lists of lists feature does not preclude #34, although it becomes less important, and it does add significantly more complexity to the processing model.

index.html
iherman: To be honest, even in OWL I don't see a way to do that.

BigBlueHat: @sgillies FWIW, we've got an issue to discuss/address that. See: #7 "support @values for describing multidimensional containers"

BigBlueHat: From a graph modeling perspective, the fact that the lat/long were originally presented in a list isn't important, is it? Only the "outer" list of coordinates has value to the graph (afaict).

azaroth42: What @gkellogg and @iherman said. In JSON-LD it will just be surface syntax of arrays. At the RDF level, it will be nested rdf:Lists with no further semantics. If more semantics are required, then a different structure would also be required.

Issue #39, opened 2018-08-01
Inclusion of @version:1.1 in documentation examples is inconsistent


gkellogg:
From json-ld/json-ld.org#652 as reported by @azaroth42 

Some examples do not have `"@context": 1.1` that still require (I think) 1.1 processing.

Example 74 (Indexing languaged-tagged strings in JSON-LD with @set representation) for example uses the 1.1 feature of @container: [@language, @set] but the context does not have the @version flag.

gkellogg:
I believe these are all fixed in PR #49.

PR #49, opened 2018-08-19
Improve example checking and generate example files

gkellogg:
This update makes use of improved features in extract-examples.rb to process all files, rather than simply doing syntactic validation, and to compare with expected results. Expected results have `@data-result-for` referencing the title of the source example. Expected results can also be in `script` elements, which don't show up in the rendered documentation.

The extract-examples script was also updated to extract JSON, as well as generate Turtle/TriG and YAML for most examples; these are updated using `rake examples`.

Eventually, multiple formats will be associated with each example, so the identification mechanism may change.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/49.html" title="Last updated on Aug 19, 2018, 12:42 AM GMT (5771ef5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/49/8530193...5771ef5.html" title="Last updated on Aug 19, 2018, 12:42 AM GMT (5771ef5)">Diff</a>


gkellogg:
Fixes #39.

PR #52, opened 2018-08-21
Update linked-data-graph …

### LINKED ISSUE #51: Images should be accessible
The spec contains a single image describing the data model ([An illustration of the data model](https://w3c.github.io/json-ld-syntax/linked-data-graph.png)), which is not accessible. Moreover, the embedded image is not scalable, and there's no description of the toolchain to recreate the PNG.

image

In CSVW, we used Keynote as the source of the images and a process (@iherman can comment) to turn images from Keynote into SVG and PNG. Images should be contained within a figure, with caption and reference to an [external description via `@longdesc`](http://w3c.github.io/csvw/metadata/properties.html). For example, from [CSVW Metadata](https://github.com/w3c/csvw/blob/gh-pages/metadata/index.html#L279-L281):

      <figure id="fig-properties" style="text-align:center">
        Metadata Properties
        <figcaption>Diagram showing the properties of different metadata descriptors (see the diagram in <a href="properties.svg">SVG</a> or <a href="properties.png">PNG</a> formats)</figcaption>
      </figure>

We should also consider other images that should be in the spec, and perhaps a simple mechanism for creating graphical representations of linked data graphs/datasets from examples.

gkellogg:
using figure and svg, with longdesc referencing an external description of the diagram. The diagram is now maintained using Google Drawings.

Fixes #51

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/52.html" title="Last updated on Aug 21, 2018, 7:30 PM GMT (360d110)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/52/cc6bad5...360d110.html" title="Last updated on Aug 21, 2018, 7:30 PM GMT (360d110)">Diff</a>


iherman:
I am fine with it now but, unfortunately, there is a strong push against longdesc which is (or will be?) removed from HTML. We may have to find another solution (e.g., using details or something similar). We may also have to look at ARIA attributes to see what can be reused from those instead.

iherman:
Yep, we can choose a technique from that document. For the WP document, we made it very simple:

https://www.w3.org/TR/wpub/#WP-diagram

But that does not scale when there are too many diagrams...

Issue #50, opened 2018-08-20
Update spec references


gkellogg:
## Multiple specs
* RFC-CONCEPTS => RFC11-CONCEPTS
* RFC2119 => RFC8174
* RFC5988 => RFC8288
* RFC6839 => RFC7303
* RFC7159 => RFC8259

## In API
ECMASCRIPT-6.0 => ECMASCRIPT – version is now 9.0, not in spec ref, but ECMASCRIPT links to it.

PR #68, opened 2018-09-23
JSON-LD in HTML

### LINKED ISSUE #23: What is 'base' for an embedded json-ld?
The definition of the [embedded JSON-LD](https://w3c.github.io/json-ld-syntax/#embedding-json-ld-in-html-documents) does not specify what the 'base' for that JSON-LD must be. It should specify it.

(cf #22)

### LINKED ISSUE #57: Make processing of embedded HTML normative
Currently, [Embedding JSON-LD in HTML Documents](https://w3c.github.io/json-ld-syntax/#embedding-json-ld-in-html-documents) is entirely informative. We've discussed making this normative, requiring JSON-LD processors to be able to identify and extract JSON-LD from a script tag with type `application/ld+json` within the HTML document.

* Given multiple such script tags, which one is used?
* Should we define a parameterized content-type to allow the version to be specified (e.g., `application/ld+json;version=1.1`)
* Does the current document base affect the base for JSON-LD processing?
  * location of  HTML document
  * `html>head>base@href`
 * `xml:base` of closest ancestor element
* Does the document language affect the default language for JSON-LD processing?
  * HTTP header- `Content-Language`
  * `@lang`, `@xml:lang`

gkellogg:
Update the JSON-LD in HTML section to be normative, describe dataset extraction, how to deal with multiple script elements and script element targeting using fragments.

I took some license to describe targeting specific script element using fragments and on the treatment of embedded escape sequences and comments.

Fixes #57.

**This should not be merged before appropriate changes made to json-ld-api.**

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/68.html" title="Last updated on Nov 16, 2018, 6:46 PM GMT (e7ae46f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/68/04fe116...e7ae46f.html" title="Last updated on Nov 16, 2018, 6:46 PM GMT (e7ae46f)">Diff</a>


index.html
```diff
@@ -8324,41 +8324,308 @@ <h3>Graph Containers</h3>
...
+           title="Embedding JSON-LD in HTML with comments">
+    <div class="selectors">
+      <button class="selected" data-selects="original">Original</button>
+      <button data-selects="expanded">Expanded</button>
+      <button data-selects="statements">Statements</button>
+      <button data-selects="turtle">Turtle</button>
+    </div>
+    <pre class="original selected" data-transform="updateExample"
+         data-content-type="text/html">
+    <!--
```
BigBlueHat: I think we'll need an example to more clearly justify this requirement. From what I've read in HTML5.2 and played with so far, it only seems to effect the parsing of the contents of a JavaScript `<script>` tag (either no `type` or one of the explicit [JavaScript MIME Types](https://www.w3.org/TR/html52/semantics-scripting.html#javascript-mime-type)).

For example, changing the [broken `<script>` example in the HTML5.2 spec](https://www.w3.org/TR/html52/semantics-scripting.html#example-8c8534f7) to use a data block, the problem goes away:
```js
<script type="application/ld+json">
  var example = 'Consider this string: <!-- <script>';
  console.log(example);
</script>
<!-- despite appearances, this is actually part of the script still! -->
<script>
  ... // this is the same script block still...
</script>
```

Consequently, I'm not sure that `<!-- -->` is ever necessary for data-blocks.

index.html
```diff
@@ -8324,41 +8324,308 @@ <h3>Graph Containers</h3>
...
+        "@vocab": "http://xmlns.com/foaf/0.1/"
+      },
+      "@id": "http://greggkellogg.net/foaf#me",
+      "@type": "Person",
+      "name": "Gregg Kellogg"
+    }
+    </script>
+
+    <p>Data described using schema.org</p>
+    <script ****id="schema"**** type="application/ld+json">
```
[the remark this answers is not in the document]
gkellogg: There are a number of places that we use both FOAF and schema.org, partly to show that similar property names can come from different vocabularies. However it's not really necessary in this case, and we could just assert different information from the same vocabulary.

index.html
BigBlueHat: So I'm working on building a [cheerio](https://cheerio.js.org/)-based data-block extractor, and I indeed did make some JSON-LD that can cause this issue. The `<!-- <script>` combo (at least) does cause it to be "overly greedy" and continue to parse beyond the intended closing tag.

I'll try and narrow in on a good/simple example, but it looks like `<!--` and `-->` handling is here to stay...

gkellogg: The text uses MAY, and you can certainly avoid using comment blocks, but that requires carefully escaping the content. So yes, it is not truly necessary to do this, but it is a common practice (which we, ourselves, use). The point is that if the content is commented, these need to be considered when processing. (I've left out CDATA, although it could be considered in an XHTML case).

The text says:

> In particular, the content MAY be enclosed in the HTML comment open (`<!--`) and comment close (`-->`) text sequences

We don't require that comments be used, but allow them to be used and place requirements on processors to consider them.

gkellogg:
Rebasing, so inline comments may be lost.

index.html
BigBlueHat: It's the "requirements on processors to consider them" that I'm "feeling" right now. 😄 Without that requirement (on implementers), you get to throw the results into `JSON.parse()` and hope for the best. With that requirement, the implementer has to do a good bit of extra effort to extract the intended contents of the `<script>`, and if not done properly essentially creates the reverse scenarios where one might extract only part of the intended data-block.

But authors come before implementers...so... 😖

gkellogg: We may need to specify steps for uncommenting and unescaping, but this seems like something we should be able to reference, and I would expect HTML libraries to handle this, but perhaps not.

BigBlueHat: HTML processors are not supposed to process any content within a [data block](https://www.w3.org/TR/html5/semantics-scripting.html#data-block):
> Setting the attribute to any other value means that the script is a data block, which is not processed. None of the script attributes (except type itself) have any effect on data blocks. Authors must use a valid MIME type that is not a JavaScript MIME type to denote data blocks.

Consequently, handling comments in the content would be up to the JSON-LD parser's processing of whatever's returned from `document.querySelector('script[type="application/ld+json"').textContent` or its corollary in a non-DOM-based parser/processor.

If it's something we expect JSON-LD processors to handle directly as input, then we'll need to define it. Otherwise, developers will have to parse out the surrounding comment markup themselves and *then* pass the results to a JSON-LD processor.

index.html
```diff
@@ -8558,10 +8560,92 @@ <h3>Graph Containers</h3>
 
   <p>When processing a JSON-LD
     <a data-cite="HTML52/semantics-scripting.html#the-script-element">script element</a>,
-    only the resolved document location of the
-    containing HTML document is used to establish the default <a>base IRI</a> of the enclosed
+    the <a data-cite="HTML52/infrastructure.html#document-base-url">Document Base URL</a>
+    of the containing HTML document,
+    as defined in [[HTML52]],
+    is used to establish the default <a>base IRI</a> of the enclosed
```
BigBlueHat: Can we put this `<base>` related content in it's own PR? I'm not confident that we have consensus as a group that this solves for the hopes/intents expressed in the [TAG's review](https://github.com/w3ctag/design-reviews/issues/312#issuecomment-434689623):
> To that end, we would like to define a JavaScript API for those script elements, to reflect that content back out to DOM.

Once there's group-wide consensus on https://github.com/w3c/json-ld-syntax/issues/23 we can revisit this prose.

gkellogg: The TAG response regarding base resolution seemed pretty clear to me:

> we would recommend that, where JSON-LD is embedded in HTML, we should treat it as part of the DOM (and it should be! A valuable part!) and resolve relative URIs to absolute URIs using the element.

However, I'm happy to extract this bit and save it for another PR. It may complicate some of the API and test changes, however.

BigBlueHat: I agree with the TAG review that JSON-LD "should be...A valuable part" of the DOM and have it's own related JS APIs to connect it within the DOM, etc. Once that's explored/"done" then I'd fully expect relative URIs to be resolved in that context.

However, as it stands it's just a data block and therefore "raw text"--which might not even be valid JSON-LD when considering it could be wrapped in HTML comments.

Consequently, we have much more to explore here (which I'm personally excited about that!). Flipping the default on how the URIs are resolved *without* doing that exploration and considering the entirety of the TAG's review would be short sighted and ultimately cause more harm than good--but this can continue on #23. 😃 

At any rate, I'd prefer things not go into the spec text until there's consensus...with apologies for it complicating the tests.

BigBlueHat: And...it seems now that this page has refreshed, you've already done that! 😁 Thanks, @gkellogg!

BigBlueHat:
Actually...before we merge this work, are we certain that the [4.12.1.3. Restrictions for contents of script elements](https://www.w3.org/TR/html5/semantics-scripting.html#restrictions-for-contents-of-script-elements) apply to [data blocks](https://www.w3.org/TR/html5/semantics-scripting.html#data-block)?

That "restrictions" section only seems to pertain to writing `<script>` content intended to be parsed by the browser's JS engine. For example, there's no reference from the "data block" examples to this "restrictions" section (and vice versa).

Thoughts?

gkellogg:
The restriction is on the contents of a script element, data blocks are a subset of all script elements. I don't follow how you infer that the section doesn't apply to data blocks. It would seem self-evident that the contents of data blocks need to have provisions for making sure that the content of the data block could be properly escaped.

In any case, these rules are necessary for properly processing JSON-LD scripts.

BigBlueHat:
@gkellogg apologies...you're right of course... I'd forgotten I'd tested all that earlier.

Toss this into http://jsbin.com/ or similar to see how it breaks:
```html
  <script type="application/ld+json" id="stuff">
  {
    "@context": {"@vocab": "http://example.com/"},
    "widget: "var example = 'Consider this string: <!-- <script>';console.log(example);"
  }
  </script>
  <script>
    <!-- whatever's below here won't run because what's above here is busted... T_T -->
    console.log(stuff.textContent);
  </script>
```

Escaping things as you've defined does fix it....sad...but true. 😩

BigBlueHat:
...hence all the red highlighting added by the Markdown parser... 🤕

gkellogg:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2018/2018-11-16-json-ld#section3).

- `Resolution #2: merge open HTML related PRs #93 and #68 and #50 after adding “At Risk” (or similar terminology) to present that things are not finalized `
What is ‘base’ for embedded json-ld?<br/>
<blockquote>
  <p><em>Ivan Herman:</em> link: <a href="https://github.com/w3c/json-ld-syntax/issues/23">https://github.com/w3c/json-ld-syntax/issues/23</a></p>
</blockquote>

<blockquote>
  <p><em>Ivan Herman:</em> Link of the PR: <a href="https://github.com/w3c/json-ld-syntax/pull/93">https://github.com/w3c/json-ld-syntax/pull/93</a></p>
</blockquote>

<blockquote>
  <p><em>Benjamin Young:</em> <a href="https://github.com/w3c/json-ld-syntax/pull/68">https://github.com/w3c/json-ld-syntax/pull/68</a></p>
</blockquote>

<blockquote>
  <p><em>Gregg Kellogg:</em> <a href="https://github.com/w3c/json-ld-api/pull/50">https://github.com/w3c/json-ld-api/pull/50</a></p>
</blockquote>

<p><strong>Benjamin Young:</strong> we discussed that one at tpac<br />
… we sent it in for TAG review, and they basically widened the scope</p>

<p><strong>Gregg Kellogg:</strong> there are 2 open PRs<br />
… 1) basic support for json-ld in html<br />
… 2) PR-93 adds text to specifically add text to add html as base<br />
… in the API spec, it’s PR-50</p>

<p><strong>Adam Soroka:</strong> quick question, what are we expected to do with their comments?<br />
… shall we respond?</p>

<p><strong>Ivan Herman:</strong> what they propose is interesting but beyond our charter<br />
… this would elevate json-ld<br />
… but yeah beyond our charter<br />
… I would say this is something the CG has to pick up<br />
… and we can cross the bridge at some time, but if this is realistic from a manpower perspective I dont know</p>

<blockquote>
  <p><em>Benjamin Young:</em> <a href="https://github.com/w3c/json-ld-syntax/pull/68#issuecomment-438015329">https://github.com/w3c/json-ld-syntax/pull/68#issuecomment-438015329</a></p>
</blockquote>

<p><strong>Ivan Herman:</strong> regarding the PR-93, there is some stuff about having XML</p>

<p><strong>Benjamin Young:</strong> the thing I just linked shows how script tags affect html parsing<br />
… it’s syntactically correct json-ld</p>

<p><strong>Gregg Kellogg:</strong> what I did in the PR-68 I call out specifics on how to handle those blocks if the media type is application/json<br />
… I think I’ve taken in the specifics on how content of script tags has to be handled and adjusted in for our needs<br />
… we asked specific questions to TAG, got an answer but they kinda got a bit over enthusiastic<br />
… out of this needs to come something that improves web platform</p>

<p><strong>Benjamin Young:</strong> the HTML comments stuff as really bothered me since I’ve read it<br />
… but it seems to primarily affect only HTML parsing<br />
… question is how much of this we need to have in the spec<br />
… json-ld in script tags vs “raw” json-ld<br />
… both have totally different escaping rules and what not.. and none of that has something to do with html base</p>

<p><strong>Ivan Herman:</strong> for the comment storing, the whole section is a normative thing<br />
… I have the impression this is an HTML problem<br />
… which we should certainly mention, but maybe not as part of a normative section</p>

<blockquote>
  <p><em>Benjamin Young:</em> HTML5 spec level text about parsing <code class="highlighter-rouge">&lt;script&gt;</code> tags <a href="https://www.w3.org/TR/html5/semantics-scripting.html#restrictions-for-contents-of-script-elements">https://www.w3.org/TR/html5/semantics-scripting.html#restrictions-for-contents-of-script-elements</a></p>
</blockquote>

<p><strong>Ivan Herman:</strong> we should officially answer to the TAG and will officially add to the standard what they said about base<br />
… and that we try to get the CG involved</p>

<blockquote>
  <p><em>Adam Soroka:</em> +1</p>
</blockquote>

<p><strong>Gregg Kellogg:</strong> comments in html and escaping.. it depends on the encoding<br />
… we don’t need to give guidance on how to handle this</p>

<blockquote>
  <p><em>Ivan Herman:</em> <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/68.html#ex-103-embedding-json-ld-in-html-with-comments">https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/68.html#ex-103-embedding-json-ld-in-html-with-comments</a></p>
</blockquote>

<p><strong>Ivan Herman:</strong> it has to be valid json-ld<br />
… that’s invalid json</p>

<p><strong>Gregg Kellogg:</strong> that’s something you see quite often</p>

<blockquote>
  <p><em>Benjamin Young:</em> ivan: because <a href="https://www.w3.org/TR/html5/semantics-scripting.html#restrictions-for-contents-of-script-elements">https://www.w3.org/TR/html5/semantics-scripting.html#restrictions-for-contents-of-script-elements</a></p>
</blockquote>

<p><strong>Gregg Kellogg:</strong> comments are often used just to make sure there are no other issues embedded in the script elements that would cause any issues</p>

<p><strong>Benjamin Young:</strong> I did quite some digging on that issue<br />
… the DOM parsing is only concerned with what’s inside the tags<br />
… it’s just treated as raw string<br />
… if there’s something inside the json-ld an html parser would choke on<br />
… the json-ld would need to be treated in such a way such that an html parser wouldn’t choke on it</p>

<p><strong>Pierre-Antoine Champin:</strong> one crazy idea by looking at the json-ld embedded in html comments: you could add a js comment in front of the html comment, making it valid javascript<br />
… it could become technically correct</p>

<p><strong>Benjamin Young:</strong> sadly it wouldn’t<br />
… it would continue being parsed<br />
… we could make it our own parsing space<br />
… the question is how far the parser gets before it finds the ending script tag</p>

<p><strong>Gregg Kellogg:</strong> the json-ld would not be allowed to contain anything that could be interpreted as html and/or html comments<br />
… not really feasible and also not helping our mission<br />
… I’ve outlined multiple approaches to tackle this<br />
… there are only a few cases where json-ld would contain things that resemble comments</p>

<p><strong>Harold Solbrig:</strong> why is this an json-ld issue but not a javascript issue?</p>

<p><strong>Gregg Kellogg:</strong> [explains why it isn’t]</p>

<p><strong>Gregg Kellogg:</strong> it did some test cases for this, exploring corner cases we know of<br />
… I don’t know how to move forward unless addressing at least some of the stuff from the PR</p>

<p><strong>Gregg Kellogg:</strong> it describes script tags and data blocks are a subset</p>

<p><strong>Benjamin Young:</strong> what’s breaking it, is the potential of one to too early close the script tag<br />
… so this would need somehow being taken care of<br />
… the risk is, the json-ld could contain content that jacks up the html it’s contained in</p>

<p><strong>Ivan Herman:</strong> is it so horrible to say, if I put json-ld in a script tag I’m supposed to escape anything that html would need to have escaped<br />
… thus a json-ld parser would have to do the unescaping<br />
… but you are in HTML regardless.. so</p>

<p><strong>Gregg Kellogg:</strong> for someone who’s actually looking at the source, those entities become rather annoying</p>

<p><strong>Ivan Herman:</strong> realistically, I don’t know how often this would happen</p>

<p><strong>Benjamin Young:</strong> the escaping issue is very similar of putting json-ld inside a text env.</p>

<p><strong>Ivan Herman:</strong> I think it’s perfectly reasonable to accept both PRs, close the issue<br />
… and open a new issue on the specific problem</p>

<p><strong>Gregg Kellogg:</strong> it’s a editor’s draft not a working draft</p>

<p><strong>Ivan Herman:</strong> we would open a issue right away</p>

<p><strong>Benjamin Young:</strong> I would only +1 this, if we add a big red AT RISK disclaimer</p>

<p><strong>Ivan Herman:</strong> a lot of very important things are pending for now<br />
… I think it’s an edge case</p>

<p><strong>Adam Soroka:</strong> I don’t think we should use a phrase like “AT RISK” but more something along the lines of “will be part of the final spec but might undergo some changes”</p>

<p><strong>Ivan Herman:</strong> we cannot commit ourselves to having always consistent editor’s drafts</p>

<p><strong>Benjamin Young:</strong> I’m not sure we have reached consensus on all the things contained</p>

<p><strong>Gregg Kellogg:</strong> I cannot work on other open issues</p>

<p><strong>Pierre-Antoine Champin:</strong> what about a parameter on the media type hinting at having to do unescaping? (like <code class="highlighter-rouge">application/ld+json;escaped=html</code>)</p>

<blockquote class="proposed_resolution">
  <p><strong>Proposed resolution: merge open HTML related PRs #93 and #68 and #50 after adding “At Risk” (or similar terminology) to present that things are not finalized</strong> <em>(Benjamin Young)</em></p>
</blockquote>

<blockquote>
  <p><em>Benjamin Young:</em> +1</p>
</blockquote>

<blockquote>
  <p><em>Adam Soroka:</em> +1</p>
</blockquote>

<p><strong>Ivan Herman:</strong> what does “that” mean?</p>

<blockquote>
  <p><em>Gregg Kellogg:</em> +1</p>
</blockquote>

<p><strong>Benjamin Young:</strong> I don’t want to have stuff merged without reaching consensus<br />
… but I could live with having it marked as being “at risk” or similar</p>

<p><strong>Ivan Herman:</strong> putting things that are already done “at risk” would be going backwards<br />
… opening a new issue that highlights things that are still being discussed would be ok<br />
… having the feeling that ~90% are done</p>

<p><strong>Adam Soroka:</strong> I have to generally agree with ivan</p>

<blockquote>
  <p><em>Ivan Herman:</em> +1</p>
</blockquote>

<p><strong>Adam Soroka:</strong> it seems for me very unlikely that we would stop talking about it</p>

<blockquote>
  <p><em>David Newbury:</em> +1</p>
</blockquote>

<blockquote>
  <p><em>Harold Solbrig:</em> +1</p>
</blockquote>

<p><strong>Benjamin Young:</strong> I’m fine with merging those</p>

<blockquote>
  <p><em>Simon Steyskal:</em> +1</p>
</blockquote>

<blockquote id="resolution2" class="resolution">
  <p><strong><em>Resolution #2: merge open HTML related PRs #93 and #68 and #50 after adding “At Risk” (or similar terminology) to present that things are not finalized</em></strong></p>
</blockquote>

<blockquote>
  <p><em>Pierre-Antoine Champin:</em> +1</p>
</blockquote>


</details>

PR #70, opened 2018-09-24
Add playground links to most examples

### LINKED ISSUE #48: JSON-LD Playground links from examples
It would be super nice to have JSON-LD Playground links available alongside any example which is itself complete JSON-LD (and doesn't contain elided sections/contents).

If this is something of interest to folks, I'm happy to add them "by hand" or look at automating the addition of the links (though I think that the elided scenarios might break that approach).

Anyhow. I think it'd be useful. 😄

gkellogg:
Only does expansion, at least for now.

Fixes #48.


index.html
```diff
@@ -749,6 +762,7 @@ <h2>The Context</h2>
         <button data-selects="expanded">Expanded</button>
         <button data-selects="statements">Statements</button>
         <button data-selects="turtle">Turtle</button>
+      <a class="playground" target="_blank"></a>
```
davidlehn: indent

index.html
```diff
@@ -2892,6 +2928,7 @@ <h3>Using the Document Base as the Default Vocabulary</h3>
     <button data-selects="expanded">Expanded</button>
     <button data-selects="statements">Statements</button>
     <button data-selects="turtle">Turtle</button>
+      <a class="playground" target="_blank"></a>
```
davidlehn: unindent

index.html
```diff
@@ -4362,6 +4411,7 @@ <h3>Using the Document Base as the Default Vocabulary</h3>
       <button data-selects="expanded">Expanded</button>
       <button data-selects="statements">Statements</button>
       <button data-selects="turtle">Turtle</button>
+    <a class="playground" target="_blank"></a>
```
davidlehn: indent

index.html
```diff
@@ -4459,6 +4509,7 @@ <h3>Using the Document Base as the Default Vocabulary</h3>
...
```
davidlehn: indent

index.html
gkellogg: Fixed.

index.html
gkellogg: Fixed.

index.html
gkellogg: Fixed.

index.html
gkellogg: Fixed.

iherman:
I wonder what the problem is, there may be some tricky HTML issues...

- The preview does not work. Unfortunately, the service is not maintained by our systeam, and Toby, who originally did it, does not have time for maintenance either
- I am unable to look at this via rawgit. I tried (both with cdn version and without), but the URL I get does not work.

Any idea?

(I am, of course, all in favor of adding those links to the examples!)

iherman:
Forget my previous remark about rawgit. It works if one uses the URL of the index.html file (not with the repo as a whole):

https://rawgit.com/w3c/json-ld-syntax/playground-links/index.html

gkellogg:
Yes, I think I know the source of the invalid JSON-LD.

Would you see a similarly styled button for this instead, or just a link such as “open in playground”?

iherman:
@gkellogg I would just use text as for all the other options

gkellogg:
@iherman Changed style to use buttons.

iherman:
Thanks! As far as I am concerned, we can merge this...

davidlehn:
It is a bit odd to have tabs you can flip between and another "tab" that is actually an action to open a another site.  Fine for now unless anyone has a better UI idea.

BigBlueHat:
Agreed with @davidlehn about needing consistency among the experience of the tabs. Also, the styles have some issues in the latest Firefox, fwiw.

gkellogg:
A UI designer I am not. I would appreciate anyone’s help in styling. 

As for the separate button for the playground link, this is entirely manipulated in JavaScript, so something else would be easy to propagate. Personally, I found the previous icon taken from json-ld.org reasonable, is it is representative of where the link will take you. When saved, there is an aria label that indicates it’s purpose.

BigBlueHat:
I missed the logo preview, but that does sound nice. Also, I sadly can't take a look at the UI stuff until next week (currently tied up with [event running](http://restfest.org/)), but I'd be happy to then. UX was most of what I did until I got into standards. 😄 🎨

gkellogg:
@BigBlueHat It's never too late on GitHub! https://rawgit.com/w3c/json-ld-syntax/7abb999a4ae7dd1b8f88d0fedb7617e16e6091fa/index.html.

PR #74, opened 2018-10-01
Type as set

gkellogg:
Allows `@type` to be defined with `@container: @set`. Most descriptions of behavior left to the API document.


gkellogg:
See also w3c/json-ld-api#41.

index.html
```diff
@@ -9008,6 +9029,9 @@ <h2>Context Definitions</h2>
     its value MUST be an <a>absolute IRI</a>, a <a>relative IRI</a>,
     or <a>null</a>.</p>
 
+  <p class="changed">If the <a>context definition</a> has an <code>@type</code> key,
+    its value MUST be a <a>dictionary</a> with the single member <code>@context</code> with value <code>@set</code>.</p>
```
davidlehn: Should be `@container`.  Probably should also be `with a value of` or `with the value`.

davidlehn:
Adding an example sure causes a gigantic mess for an otherwise small change.

index.html
davidlehn: Or maybe `<code>@container</code> set to <code>@set</code>` to align with the api text?

PR #90, opened 2018-10-31
Mark blank node property labels as obsolete

### LINKED ISSUE #37: Consider obsoleting use of blank nodes for properties and "generalized RDF"
This is one of the major things that makes JSON-LD out of step with the RDF data model, and it's not clear if the feature is used or valuable. IIRC, the original issue was making it easy to support mapping ad-hoc JSON structures without creating IRIs, but the use of `@vocab`, and document-relative IRIs for properties would seem to obviate the need for this.

This would likely prohibit the direct use of blank node identifiers in the property position, as well as the mapping of terms to blank node identifiers.

Downsides: another area of potential incompatibility with JSON-LD 1.0. A backwards-compatible solution would be to preserve the feature, but mark it archaic. This might cause warnings to be generated if encountered with a processor running in 1.1 mode.

gkellogg:
…and generalized RDF datasets.

Fixes #37.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9187,7 +9187,7 @@ the data type to be specified explicitly with each piece of data.</p>
   </ul>
 
   <p class="issue atrisk">The use of <a>blank node identifiers</a> to label properties is obsolete,
-    and may be removed in a future version of JSON-LD, as is the support for <a>generalized rdf datasets</a>.</p>
+    and may be removed in a future version of JSON-LD, as is the support for <a>generalized RDF Datasets</a>.</p>
 
   <p>Summarized, these differences mean that JSON-LD is capable of serializing any RDF
     <a>graph</a> or <a>dataset</a> and most, but not all, JSON-LD documents can be directly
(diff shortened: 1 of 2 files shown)


iherman:
Damn, the preview tool does not work:-(

index.html
```diff
@@ -9183,33 +9186,20 @@ <h2>Relationship to RDF</h2>
       between JSON's native data types and RDF's counterparts to allow round-tripping.</li>
   </ul>
 
+  <p class="issue atrisk">The use of <a>blank node identifiers</a> to label properties is obsolete,
+    and may be removed in a future version of JSON-LD, as is the support for <a>generalized rdf datasets</a>.</p>
```
davidlehn: rdf -> RDF

index.html
```diff
@@ -9895,6 +9885,8 @@ <h2>Changes since JSON-LD Community Group Final Report</h2>
       to ensure that <code>@type</code> members are always represented as an array. This
       also allows a term to be defined for <code>@type</code>, where the value MUST be a <a>dictionary</a>
       with <code>@container</code> set to <code>@set</code>.</li>
+    <li>The use of <a>blank node identifiers</a> to label properties is obsolete,
+      and may be removed in a future version of JSON-LD, as is the support for <a>generalized rdf datasets</a>.</li>
```
davidlehn: rdf -> RDF

gkellogg:
@iherman I don't know what to say about the lake of the Preview/Diffs. Last time it was because of a problem with spec-gen, but it seems to be running properly now. I agree that it's frustrating when it simply doesn't show up. Same problem in w3c/json-ld-api#49.

PR #92, opened 2018-11-15
Improve keyword descriptions

### LINKED ISSUE #77: Disambiguate uses of @type
Related to json-ld-api#33, we discussed on 2018-10-05 WG call the distinction between `@type` as a baked in alias for `rdf:type`, and `@type` as a function of the syntax for specifying the data type of the JSON object (resource vs literal, and type of literal), via the context.

There was general agreement that this was a common source of confusion, and could be ameliorated for expansion by introducing an alias (for example `@datatype`) but that this would add to confusion in compaction, as it would replace `@type` everywhere.  A more robust change would require a change to the specification to allow either `@type` or `@datatype` as the way to specify the data type, however that is a significant addition that should be considered in detail.

Further, one might introduce `type`: `rdf:type` in a context, rather than use the more magical `@type`. This approach would also impact the specifications, but would follow from the rejection of the `@label` syntactic sugar  in #6.

gkellogg:
* Improved `@type` keyword definition, and added _Uses of JSON Objects_ section.
* Add a section on keywords to the grammar, and slight improvement to type values.

Addresses #77.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/92.html" title="Last updated on Nov 17, 2018, 7:27 PM GMT (a42ebb3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/92/337f184...a42ebb3.html" title="Last updated on Nov 17, 2018, 7:27 PM GMT (a42ebb3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -494,13 +494,13 @@
       <dd>Used to set the type of a <a>node</a> or the datatype of a <a>typed value</a>.
         This keyword is described further in <a class="sectionRef" href="#specifying-the-type"></a>
         and <a class="sectionRef" href="#typed-values"></a>.
-        <div class="note">The notion of type can be confusing, as both
-          nodes and literal values can have types,
-          but in the <a>data model</a>,
-          these are fundamentally different things.
-          The use of <code>@type</code> to define a type for both <a>node objects</a> and <a>value objects</a>
-          addresses the basic need to type data,
-          be it a literal value or a more complicated <a>resource</a>.
+        <div class="note">The use of <code>@type</code> to define a type for both
+          <a>node objects</a> and <a>value objects</a> addresses the basic need to type data,
+          be it a literal value or a more complicated resource.
+          Experts may find the overloaded use of the <code>@type</code> keyword for both purposes concerning,
+          but should note that Web developer usage of this feature over multiple years
+          has not resulted in its misuse due to the far less frequent use of <code>@type</code>
+          to express typed literal values.
         </div>
       </dd>
       <dt><code>@container</code></dt>
@@ -1453,7 +1453,7 @@
     <li><a>JSON objects</a>, which provide a set of <a>dictionary members</a>, relating keys with values.</li>
   </ul>
 
-  <p>The JSON-LD data model allows for a richer set of resources, based on the RDF data mode.
+  <p>The JSON-LD data model allows for a richer set of resources, based on the RDF data model.
     The data model is described more fully in <a class="sectionRef" href="#data-model"></a>.
     JSON-LD uses JSON objects to describe various resources, along with the relationships
     between these resources:</p>
@@ -3114,7 +3114,7 @@ type with a particular <a>term</a> in the <code>@context</code>:</p>
   </pre>
 </aside>
 
-<p>The <em>modified</em> key's value above is automatically type interpreted as a
+<p>The <em>modified</em> key's value above is automatically interpreted as a
   <em>dateTime</em> value because of the information specified in the
   <code>@context</code>. The example tabs show how a <a>JSON-LD processor</a> will interpret the data.</p>
 
@@ -9643,24 +9643,24 @@ the data type to be specified explicitly with each piece of data.</p>
     <dt><code>@container</code></dt><dd>
       The <code>@container</code> keyword MUST NOT be aliased, and MAY be used as a key in an <a>expanded term definition</a>.
       Its value MUST be either
-          <code>@list</code>,
-          <code>@set</code>,
-          <code>@language</code>,
-          <code>@index</code>,
-          <span><code>@id</code></span>,
-          <span><code>@graph</code></span>,
-          <span><code>@type</code></span>, or be
-          <a>null</a>,
-          or an <a>array</a> containing exactly any one of those keywords, or a
-          combination of <code>@set</code> and any of <code>@index</code>,
-          <code>@id</code>, <code>@graph</code>, <code>@type</code>,
-          <code>@language</code> in any order.
-          The value may also be an array
-          containing <code>@graph</code> along with either <code>@id</code> or
-          <code>@index</code> and also optionally including <code>@set</code>.</span>
+      <code>@list</code>,
+      <code>@set</code>,
+      <code>@language</code>,
+      <code>@index</code>,
+      <span><code>@id</code></span>,
+      <span><code>@graph</code></span>,
+      <span><code>@type</code></span>, or be
+      <a>null</a>,
+      or an <a>array</a> containing exactly any one of those keywords, or a
+      combination of <code>@set</code> and any of <code>@index</code>,
+      <code>@id</code>, <code>@graph</code>, <code>@type</code>,
+      <code>@language</code> in any order.
+      The value may also be an array
+      containing <code>@graph</code> along with either <code>@id</code> or
+      <code>@index</code> and also optionally including <code>@set</code>.
     </dd>
     <dt><code>@context</code></dt><dd>
-      The <code>@context</code> keyword MUST NOT be aliased, and MAY be used as a key in the following objects:</p>
+      The <code>@context</code> keyword MUST NOT be aliased, and MAY be used as a key in the following objects:
       <ul data-sort>
         <li><a>node objects</a> (see <a class="sectionRef" href="#node-objects"></a>),</li>
         <li><a>value objects</a> (see <a class="sectionRef" href="#value-objects"></a>),</li>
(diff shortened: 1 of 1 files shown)


gkellogg:
@azaroth42 I didn't see a way to significantly improve 4.2.1; perhaps the other changes do the job.

index.html
```diff
@@ -1431,6 +1440,86 @@ <h2>Specifying the Type</h2>
...
+  <ul>
+    <li><a>Numbers</a>, which describe literal numeric values,</li>
+    <li><a>Strings</a>, which may describe literal string values, or be used as the keys in a <a>JSON object</a>.</li>
+    <li><a>Boolean</a> <code>true</code> and <code>false</code>, which describe literal boolean values,</li>
+    <li><code>Null</code>, which describes the absense of a value,</li>
+    <li><a>Arrays</a>, which describe an ordered set of values of any type, and</li>
+    <li><a>JSON objects</a>, which provide a set of <a>dictionary members</a>, relating keys with values.</li>
+  </ul>
+
+  <p>The JSON-LD data model allows for a richer set of resources, based on the RDF data mode.
```
msporny: Typo "mode" -> "model"

index.html
```diff
@@ -3025,7 +3114,7 @@ <h3>Using the Document Base as the Default Vocabulary</h3>
   </pre>
 </aside>
 
-<p>The <em>modified</em> key's value above is automatically type coerced to a
+<p>The <em>modified</em> key's value above is automatically type interpreted as a
```
msporny: "type interpreted"? not familiar with that term? Did we rename "type coercion" to "type interpretation"?

index.html
```diff
@@ -9153,6 +9243,175 @@ <h2>Context Definitions</h2>
...
+        <li><a>node object</a>,</li>
+        <li><a>value object</a>, or</li>
+        <li>an <a>array</a> of zero or more of the above possibilities</li>
+      </ul>
+
+      <p>See <a class="sectionRef" href="#sets-and-lists"></a> for further discussion on sets and lists.</p>
+    </dd>
+    <dt><code>@type</code></dt><dd>
+      The <code>@type</code> keyword MAY be aliased and MAY be used as a key in a <a>node object</a> or a <a>value object</a>.
+      The unaliased <code>@type</code> MAY be used as a key in an <a>expanded term definition</a>,
```
msporny: This makes it sound like an aliased `@type` MAY be used as a key in an expanded term definition... guessing other rules prevent this... probably don't need to change it, but my reading of the previous sentence and this sentence threw me a bit.

index.html
```diff
@@ -491,9 +491,18 @@ <h2>Syntax Tokens and Keywords</h2>
         language of a JSON-LD document. This keyword is described in
         <a class="sectionRef" href="#string-internationalization"></a>.</dd>
       <dt><code>@type</code></dt>
-      <dd>Used to set the data type of a <a>node</a> or
-        <a>typed value</a>. This keyword is described in
-        <a class="sectionRef" href="#typed-values"></a>.</dd>
+      <dd>Used to set the type of a <a>node</a> or the datatype of a <a>typed value</a>.
+        This keyword is described further in <a class="sectionRef" href="#specifying-the-type"></a>
+        and <a class="sectionRef" href="#typed-values"></a>.
+        <div class="note">The notion of type can be confusing, as both
```
msporny: I wouldn't start by saying that this can be confusing. It's confusing to people that have deep knowledge, but for beginners I don't expect it is. 

Things have types. Nodes have types. Values have types. You use `@type` to express that. That's not confusing. 

The confusing part is probably the *chording* of `@type` with other things like `@value` and a node to get the expression you want... but we really only do that in one place that only experts tend to see.

In any case, I'd reword to something like:

The use of <code>@type</code> to define a type for both <a>node objects</a> and <a>value objects</a>        addresses the basic need to type data, be it a literal value or a more complicated <a>resource</a>. Experts may find the overloaded use of the `@type` keyword for both purposes concerning, but should note that Web developer usage of this feature over multiple years has not resulted in its misuse due to the far less frequent use of `@type` to express typed literal values.

iherman:
@gkellogg I like what I see, but I have two more proposals

- I think it may be worth exchanging the order for 3.4 and 3.5. The new 3.5 could then very explicitly refer to the fact that `@type` is used to set a type on value objects and node objects, something that will have been described in the previous section. I would think that 3.5 should then have two subsections: one for node objects (which is the current content, essentially) and one for value objects, describing the usage of datatypes. These two sections together might then settle the original issue.
- You use the term map objects, which is fine, but this may just indicate that the term `@container` to denote all these objects may not be the right term and may be another source of confusion. We should keep the `@container` for lists or sets (and, of course, we should allow the old usage for, eg, language maps) but we should have a separate `@map` to define, well, the map objects. We should discuss that separately, though.

index.html
gkellogg: @iherman had some pushback on the notion of "type coercion", and "type interpretation" seemed like it might be a better fit. What happens in this case, is that the `@type` in the term definition is what allows the string value to be interpreted as a `dateTime`. But, it should probably be "interpreted as" rather than "type interpreted as".

Personally, I could go either way on this. Strictly speaking, "coercion" would be when something of one type is converted to another type, perhaps automatically because of an operation such as addition or concatenation. In this case, arguably, the string has no inherent type, and it is interpreted as an `xsd:string`, in the absence of any more explicit `@type` declaration.

index.html
gkellogg: This is a pattern I came up with, and I'm happy to do something else. The first sentence says that it may be aliased when used in a **node object** or **value object**. The second is that it may be used unaliased as a key in an **expanded term definition**.

Suggestions for a clearer description welcome.

index.html
gkellogg: But, the fact is, for many it seems to be a confusing issue. But, I like your wording better!

gkellogg:
@iherman Your first point is well taken, and I'll make the suggested change.

Regarding changing `@container` to `@map` for new map objects, I disagree. The mold was cast when we used `@container` for language- and data-maps. Also, it would add additional steps to the context and expansion algorithms, to no great effect. I thought we had already discussed this on a call.

gkellogg:
@iherman I made  your suggested changes to "Use of JSON Objects" and "Specifying the Type".

iherman:
@gkellogg thanks for the changes you made. Two more comments (just for nagging you:-)

- I think adding an example for ~datatype~, ehem, value object type, would be good as well. Just to have this on an equal footing, so to way, with the node object type; it currently reads as some sort of a side remark.
- the second note in that section repeats information that is now part of the core text, and should be removed.

Thx!

gkellogg:
> I think adding an example for datatype, ehem, value object type, would be good as well. Just to have this on an equal footing, so to way, with the node object type; it currently reads as some sort of a side remark.

There are examples in the referenced sections, I don't see a point in repeating them, particularly as the section is mostly about typed objects. The "Specifying the Type" section is really about typing node objects, to make it clear, I discuss typing value objects and reference "Typed Values" and "Type Coercion". Do you really think we need to repeat that detail here?

The reason "Specifying the Type" is in the "Basic Concepts" section, as this is considered a basic concept, while typed values are considered advanced. Otherwise, we'd need a separate section "Describing Nodes". But, there's really not enough there do do that, IMHO.

> the second note in that section repeats information that is now part of the core text, and should be removed.

I'll pare out some of the repeated detail, but the original purpose of the note is that this describes basic features. Instead, I'll reference the "Advanced Concepts" section.

iherman:
@gkellogg first of all, I do not want to micromanage. As far as I am concerned this PR can be merged right away and it may be indeed enough to answer to #77. You are free to take or drop any of my comments.

That being said, I guess the main reason of my comment is this remark:

> The reason "Specifying the Type" is in the "Basic Concepts" section, as this is considered a basic concept, while typed values are considered advanced. Otherwise, we'd need a separate section "Describing Nodes". 

I do not feel that typing a node is a basic concept whereas typing a value is an advanced. To be more precise, I do not see why these two should be different. You are right that section 4.2 is already there, in particular 4.2.1, and I understand that you do not want to repeat things. My feeling is, however, that 4.2 and 3.5 should be on equal footing; I do not have a strong feeling whether 3.5 should be moved into the 'advanced' category, or 4.2. (or possibly 4.2.1. only) should be put into the basic section, but I would think these two should be somehow kept together.

Again, regardless from this, 👍 from me to merge at your discretion.

PR #106, opened 2018-12-08
Make requirement for entity encoding non-normative

### LINKED ISSUE #100: Handling HTML character references in embedded JSON-LD
PR #97 and w3c/json-ld-api#51 resolved embedding JSON-LD in provide for HTML character references (entities) to be encoded within a data block so that critical elements are not mishandled by HTML parsers. As written this requires authors to encode sequences such as `<script></script>` which may occur in JSON-LD strings, and for these to be decoded when being extracted by the API.

It seems that the structured-data testing tool does not do this, and it's not clear what provisions are made for potentially problematic content.

One way to resolve this would be to provide direction to content authors to avoid any JSON-LD content embedded in HTML which could cause such a problem, which seems to be current practice. If character references are encoded, decoding is required at some layer.

cc/ @danbri @BigBlueHat

gkellogg:
and update examples to not show automatic decoding.

Fixes #100.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/106.html" title="Last updated on Dec 8, 2018, 7:50 PM GMT (b12a965)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/106/28df8b5...b12a965.html" title="Last updated on Dec 8, 2018, 7:50 PM GMT (b12a965)">Diff</a>

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,11 +1,11 @@
 GIT
   remote: git://github.com/ruby-rdf/json-ld.git
-  revision: 488d5e7737d401bb83a246248ad2f921c56938c4
+  revision: a26c3205d006550f85162848a4a2bc3366b39970
   branch: develop
   specs:
     json-ld (3.0.2)
-      multi_json (~> 1.12)
-      rdf (>= 2.2.8, < 4.0)
+      multi_json (~> 1.13)
+      rdf (~> 3.0, >= 3.0.4)
 
 GEM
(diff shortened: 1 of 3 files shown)


gkellogg:
@danbri We resolved to remove decoding of HTML character references, which is consistent with how the SDTT handles it.

PR #110, opened 2018-12-18
Added informative reference to SPARQL11-OVERVIEW in intro

pchampin:
@gkellogg I'd rather use a PR for the moment, to be sure I don't mess things up.

You'll have to explain to me what magic inserted the correct text for SPARQL11-OVERVIEW in the reference section -- and what would have been required to insert it in the *normative* references instead...

### DIFF
--- a/index.html
+++ b/index.html
@@ -284,7 +284,7 @@
   <p>
    JSON-LD is designed to be usable directly as JSON, with no knowledge of RDF
    [[RDF11-CONCEPTS]]. It is also designed to be usable as RDF, if desired, for
-   use with other Linked Data technologies like SPARQL. Developers who
+   use with other Linked Data technologies like SPARQL [[SPARQL11-OVERVIEW]]. Developers who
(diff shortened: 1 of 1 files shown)


gkellogg:
@iherman will need to figure out the IPR issues.

Regarding normative/informative references, ReSpec recently added some magic to figure this out, based on if it’s in a normative or informative section. It can be overridden, but the details of this are on the ReSpec wiki.

index.html
```diff
@@ -284,7 +284,7 @@ <h1>Introduction</h1>
   <p>
    JSON-LD is designed to be usable directly as JSON, with no knowledge of RDF
    [[RDF11-CONCEPTS]]. It is also designed to be usable as RDF, if desired, for
-   use with other Linked Data technologies like SPARQL. Developers who
+   use with other Linked Data technologies like SPARQL [[SPARQL11-OVERVIEW]]. Developers who
```
gkellogg: I believe `[[!SPARQ=OVERVIEW]]` would be normative.

PR #126, opened 2019-02-05
Clarify compact IRI term example

### LINKED ISSUE #123: IRI expansions in the context... a bit messy:-(
While re-reading the syntax document I got to [example 36](https://w3c.github.io/json-ld-syntax/#example-36-using-a-compact-iri-as-a-term) and [example 37](https://w3c.github.io/json-ld-syntax/#example-37-associating-context-definitions-with-absolute-iris) and I must admit it is still a mystery to me. I wonder whether this is purely editorial and can be explained better or whether it _is_ messy.

If the latter, we may want to consider revising this part of the algorithm and possibly disallowing certain combination. As it stands, this is the type of complication that may drive away people from JSON-LD... ☹️

gkellogg:
Make example 36 "Using a compact IRI as a term" clearer, but only defining `foaf:homepage` using the compact IRI term form.

Fixes #123.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/126.html" title="Last updated on Feb 12, 2019, 11:22 PM UTC (bc3a64c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/126/1533ca9...bc3a64c.html" title="Last updated on Feb 12, 2019, 11:22 PM UTC (bc3a64c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2816,8 +2816,8 @@
     ****"foaf": "http://xmlns.com/foaf/0.1/"****,
     "xsd": "http://www.w3.org/2001/XMLSchema#",
     "name": "foaf:name",
-    "age": {
-      "@id": "foaf:age",
+    ""****foaf:age"****": {
+      "@id": "****http://xmlns.com/foaf/0.1/age****",
       "@type": "xsd:integer"
     },
     "****foaf:homepage****": ****{
--- a/index.html
+++ b/index.html
@@ -2816,7 +2816,8 @@
     ****"foaf": "http://xmlns.com/foaf/0.1/"****,
     "xsd": "http://www.w3.org/2001/XMLSchema#",
     "name": "foaf:name",
-    "****foaf:age****": {
+    "age": {
+      "@id": "foaf:age",
       "@type": "xsd:integer"
     },
     "****foaf:homepage****": ****{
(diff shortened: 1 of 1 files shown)


pchampin:
I think that the text after the example would also require some adaptation.

* It says "`foaf:age` declares both the IRI for the term (using short-form) as well as the `@type` associated with the term". Wouldn't it be more accurate now to state that "`age` declares both..."?

* It says "The compact IRI form is used in two different ways", while there are actually 4 compact IRIs (`foaf:name` and `xsd:integer` are also used). Before, I assumed that it meant "The compact IRI form *on the left-hand side*", but this is obviously not the case anymore... In fact, since the term definition for age is now the same as in the previous examples, I would not mention it at all, and focus solely on the `foaf:homepage` rule, which is what makes this example distinctive.

* Finally, I noticed that, in example 37 (the next one), we have
```
  ...
  "foaf:age": {
    "@id": "foaf:age",
    ...
  }
```
which is correct but quite redundant. For the sake of clarity (and continuity in the examples) I suggest replacing `foaf:age` by `age`.

gkellogg:
To be consistent with the description, and intent, I went back to using `foaf:age` as the term, but changed its `@id` to be an absolute IRI.

PR #131, opened 2019-02-19
Expand compaction section

gkellogg:
…different term definitions on compaction. This includes describing `@type: @none`.

For w3c/json-ld-api#33.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/131.html" title="Last updated on Feb 20, 2019, 6:38 PM UTC (9c1cbcc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/131/f08ced8...9c1cbcc.html" title="Last updated on Feb 20, 2019, 6:38 PM UTC (9c1cbcc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -9357,10 +9357,10 @@ the data type to be specified explicitly with each piece of data.</p>
     <p>Generally, when compacting, properties having only one value are
       represented as strings or dictionaries, while properties having
       multiple values are represented as an array of strings or dictionaries.
-      This means that applications accessing such properties needs to be prepared
+      This means that applications accessing such properties need to be prepared
       to accept either representation. To force all values to be represented
       using an array, a term definition can set <code>"@container": "@set"</code>.
-      Moreover, `@set` can be used in combination with other container settings,
+      Moreover, <code>@set</code> can be used in combination with other container settings,
       for example looking at our language-map example from <a href="#indexing-values" class="sectionRef"></a>:</p>
 
     <aside class="example" title="Indexing language-tagged strings and @set">
--- a/index.html
+++ b/index.html
@@ -9035,7 +9035,7 @@ the data type to be specified explicitly with each piece of data.</p>
         {"@value": "English", "@language": "en"}
       ],
       "http://example.com/iri": [
-        {"@id": "http://example.org/iri"}
+        {"@id": "http://example.com/some-location"}
       ]
     }]
     -->
@@ -9076,7 +9076,7 @@ the data type to be specified explicitly with each piece of data.</p>
       "plain": ["string", true, 1],
       "date": "2018-02-16",
       "en": "English",
-      "iri": "http://example.org/iri"
+      "iri": "http://example.com/some-location"
     }
     -->
     </pre>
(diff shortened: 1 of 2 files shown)


gkellogg:
Work is incomplete for more descriptions of compaction, but the bit on `@type: @none` is in. Please comment.

iherman:
Mini-mini editorial comment: it would be good to use a non-breaking space character between a "§" and the number. I get lines with a "§" at the end of the line and the number at the beginning of the next line; it does look strange...

iherman:
Just an aesthetic issue... In example 105 you say:

```
"http://example.com/iri": [
    {"@id": "http://example.org/iri"}
  ]
```

Maybe it is nicer to use a different URI for the term and its value :-)

iherman:
In example 106 you say:

> Note that there is no term defined for "plain", that is created automatically using the vocabulary mapping. 

I am not sure what this refers to. In the example, "plain" is a simple text, and stays that way. Is this a leftover from a different example?

gkellogg:
> Mini-mini editorial comment: it would be good to use a non-breaking space character between a "§" and the number. I get lines with a "§" at the end of the line and the number at the beginning of the next line; it does look strange...

That would be an issue for ReSpec, which added the "§” recently.

iherman:
> That would be an issue for ReSpec, which added the "§” recently.

Ah. I have raised https://github.com/w3c/respec/issues/2109.

index.html
```diff
@@ -8887,8 +8887,292 @@ <h3>Graph Containers</h3>
...
+        "notype": {"@id": "http://example.com/notype", "@type": "@none"}
+      },
+      "notype": [
+        {"@value": "string"},
+        {"@value": true},
+        {"@value": false},
+        {"@value": 1},
+        {"@value": 10.0},
+        {"@value": "plain"},
+        {"@value": true, "@type": "xsd:boolean"},
```
pchampin: shouldn't that be `{"@value": "true", "@type": "xsd:boolean"},` (with quotes around `true`)? I get that both are valid, but I find it odd to use a native value with an explicit datatype...

gkellogg: It certainly has the same meaning, but value compaction doesn't have a case for this; If the term were defined with `@type: xsd:boolean` it would certainly collapse down to just `true`.

(To make this a tad less confusing, as we already have `{"@value": true}`, I'm using `{"@value": false, "@type": "http://www.w3.org/2001/XMLSchema#boolean"}` instead).

We could update value compaction for native values (although, it can get tricky for something like `xsd:decimal`), but we'd risk a backwards incompatible change.

It may be simpler to just leave out this particular value for this example.

pchampin: I'm not suggesting that compaction convert `true` to `"true"` in that case... only that the *source* use a string (and so compaction gives that same string back). Unless you think that it is acceptable practice to author JSON-LD with non-strings in the `"@value"` field of a value node... I still find it odd but I can live with that.

> To make this a tad less confusing...

Good idea :) I didn't notice that, but it *can* be confusing :)

gkellogg: Compaction should not change native values (such as `true` or `false` to strings), and the `fromRdf` algorithm has an option to use native values when possible. Because of this, when expanding something like `"term": true`, it will always using the value object representation of `"http://example.org/term": [{"@value": true}]`.

We could certainly update the example to use "true" (or "false") instead of the native `true`, which may make this less confusing.

PR #133, opened 2019-02-20
Update implicit named graph example.

### LINKED ISSUE #122: Graph container text and implementation do not match...
The text underneath https://w3c.github.io/json-ld-syntax/#ex-88-implicitly-named-graph says:

> The example above expresses a named graph that is identified by the blank node identifier _:b0. That graph is composed of the statements about Manu and Gregg. Metadata about the graph itself is expressed via the generatedAt property, which specifies when the graph was generated.

This seems to be wrong. The playground result, as well as the text above the example suggests that it should say something like:

> The example above expresses two named graphs, identified by blank nodes. These graphs are composed of the statements about Manu and Gregg, respectively, each in its own separate graphs. 

Strictly speaking, the `generatedAt` is about a _resource_ (`http://example.org/foaf-graph`) which is not a dataset/graph, "just" a resource.

We may want to spend some time looking over the whole section...

gkellogg:
Note, this will need to be updated when vc-data-model changes and their context is updated. Liberties were taken.

Fixes #122.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/133.html" title="Last updated on Feb 22, 2019, 12:15 AM UTC (f9ef6c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/133/f08ced8...f9ef6c9.html" title="Last updated on Feb 22, 2019, 12:15 AM UTC (f9ef6c9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -8019,13 +8019,13 @@ the data type to be specified explicitly with each piece of data.</p>
       {
         "@context": {
           "@version": 1.1,
-         "@base": "http://dbpedia.org/resource/",
+          "@base": "http://dbpedia.org/resource/",
           "id": "@id",
           "isa": "@type",
-          "said": {"@id": "http://example.com/said", "@container": "@graph"}
+          ****"said": {"@id": "http://example.com/said", "@container": "@graph"}****
         },
         "id": "Epimenides",
-        "said": {
+        ****"said"****: {
           "id": "Epimenides",
           "isa": "Liar"
         }
@@ -8037,8 +8037,8 @@ the data type to be specified explicitly with each piece of data.</p>
       <!--
       [{
         "@id": "http://dbpedia.org/resource/Epimenides",
-        "http://example.com/said": [{
-          "@graph": [{
+        ****"http://example.com/said"****: [{
+          ****"@graph"****: [{
             "@id": "http://dbpedia.org/resource/Epimenides",
             "@type": ["http://dbpedia.org/resource/Liar"]
           }]
--- a/index.html
+++ b/index.html
@@ -8020,14 +8020,13 @@ the data type to be specified explicitly with each piece of data.</p>
         "@context": {
           "@version": 1.1,
           "@base": "http://dbpedia.org/resource/",
-          "id": "@id",
-          "isa": "@type",
-          ****"said": {"@id": "http://example.com/said", "@container": "@graph"}****
+          "said": "http://example.com/said",
+          ****"wrote": {"@id": "http://example.com/wrote", "@container": "@graph"}****
         },
-        "id": "Epimenides",
-        ****"said"****: {
-          "id": "Epimenides",
-          "isa": "Liar"
+        "@id": "Skakespeare",
+        ****"wrote"****: {
+          "@id": "Richard_III_of_England",
+          "said": "My kingdom for a horse"
         }
       }
       -->
@@ -8036,11 +8035,11 @@ the data type to be specified explicitly with each piece of data.</p>
            data-result-for="Implicitly named graph-original">
       <!--
       [{
-        "@id": "http://dbpedia.org/resource/Epimenides",
-        ****"http://example.com/said"****: [{
+        "@id": "http://dbpedia.org/resource/Skakespeare",
+        ****"http://example.com/wrote"****: [{
           ****"@graph"****: [{
-            "@id": "http://dbpedia.org/resource/Epimenides",
-            "@type": ["http://dbpedia.org/resource/Liar"]
+            "@id": "http://dbpedia.org/resource/Richard_III_of_England",
+            "http://example.com/said": [{"@value": "My kingdom for a horse"}]
           }]
         }]
       }]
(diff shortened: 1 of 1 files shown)


gkellogg:
@dlongley Note that I took some liberties on Example 33 from vc-data-model. Once your context is updated, and published at https://www.w3.org/2018/credentials/v1, we can update this.

Also, note that the value of `issuanceDate` is out of range, you use an xsd:date, and it should be an xsd:dateTime, according to the RDFS definition.

index.html
```diff
@@ -8018,63 +8018,64 @@ <h3>Graph Containers</h3>
...
-          "@version": 1.1,
-          "generatedAt": {
-            "@id": "http://www.w3.org/ns/prov#generatedAtTime",
-            "@type": "http://www.w3.org/2001/XMLSchema#date"
+        "@context": "https://w3c.github.io/vc-data-model/vocab/credentials.jsonld",
+        "id": "http://example.edu/credentials/332",
+        "type": "VerifiableCredential",
+        "issuer": "https://example.edu/issuers/4",
+        "issuanceDate": "2017-02-24T00:00:00Z",
+        ****"verifiableCredential"****: {
```
```suggestion
        ****"credentialSubject"****: {
```

index.html
```diff
@@ -8018,63 +8018,64 @@ <h3>Graph Containers</h3>
...
+        "https://w3.org/2018/credentials#issuer": [
+          {"@id": "https://example.edu/issuers/4"}
+        ],
+        "https://w3.org/2018/credentials#issuanceDate": [
+          {
+            "@value": "2017-02-24T00:00:00Z",
+            "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+          }
+        ],
+        "https://w3.org/2018/credentials#verifiableCredential": [{
```
```suggestion
        "https://w3.org/2018/credentials#credentialSubject": [{
```

index.html
dlongley: The quads, etc. will also need to change of course.

dlongley: Note that `credentialSubject` is not an `@graph` container. The term `verifiableCredential` is one, however, it is used in a `VerifablePresentation` not a `VerifiableCredential`. That being said, `proof` *is* an `@graph` container ... and since @iherman asked for just one example of a `@graph` container we could leave this as is to address his concern. Otherwise, we should wrap the whole credential in a `VerifiablePresentation` that points to the credential via `verifiableCredential`.

gkellogg: Based on @iherman's comments, we probably need to go to an entirely different example, anyway. Or, trim down the example and use an inline context.

gkellogg:
With apologies to @msporny, I changed the example to be a different use of quoted graphs.

dlongley:
I guess I should also say that we should, IMO, to the best extent that we can, avoid using prefixed terms in JSON in examples (particularly new examples that we add). I think it's fine to do that in examples where we're showing that you *can* use the prefix feature, but it would be best to focus on more idiomatic JSON in all other examples as much as possible.

gkellogg:
> I guess I should also say that we should, IMO, to the best extent that we can, avoid using prefixed terms in JSON in examples (particularly new examples that we add). I think it's fine to do that in examples where we're showing that you _can_ use the prefix feature, but it would be best to focus on more idiomatic JSON in all other examples as much as possible.

That's easy enough to fix, but I was trying to keep the context size down.

dlongley:
Another option from Manu that is less controversial: http://manu.sporny.org/2016/json-ld-context-caching/

pchampin:
Sorry to be nitpicking, but I don't really like the MLK example either... I don't see it as a compelling use case for a named graph. What exactly does `_:b2 :quote "I have a dream"` mean? Why not say `:mlk :said "I have a dream"`?

Here is a counter-proposal:
```
{
  "@context": {
    "@version": 1.1,
    "said": {"@id": "http://example.com/said", "@container": "@graph"},
    "Liar": "http://example.com/Liar"
  },
  "@id": "http://dbpedia.org/resource/Epimenides",
  "said": {
    "@id": "http://dbpedia.org/resource/Epimenides",
    "@type": "Liar"
  }
}
```

Granted, it is less idiomatic (explicit use of `@id` and `@type`). but this could be fixed with a heavier context...

```
{
  "@context": {
    "@version": 1.1,
    "id": "@id",
    "isa": "@type",
    "Epimenides": "http://dbpedia.org/resource/Epimenides",
    "said": {"@id": "http://example.com/said", "@container": "@graph"},
    "Liar": "http://example.com/Liar"
  },
  "id": "Epimenides",
  "said": {
    "id": "Epimenides",
    "isa": "Liar"
  }
}
```

gkellogg:
Actually, terms don't work as a value of `@id`, so it might need to be something like:

```json
{
  "@context": {
    "@version": 1.1,
   "@base": "http://dbpedia.org/resource/",
    "id": "@id",
    "isa": "@type",
    "said": {"@id": "http://example.com/said", "@container": "@graph"}
  },
  "id": "Epimenides",
  "said": {
    "id": "Epimenides",
    "isa": "Liar"
  }
}
```

gkellogg:
Why I think the Epimenides example may be better is that it uses a contradiction, for which a named graph seems appropriate. Otherwise, why use a named graph at all?

Let's leave it until tomorrow.

pchampin:
More than the contradiction, I think it is the self-reference that's
bothering @dlongley.

Here's another example :

Skakespeare said that Richard III said "My kingdom for a horse".

iherman:
An aside... I know many use the `{"id":"@id"}` in the contexts, including myself. However, when we have an example in the document, we should try to keep only those statements that have an importance in understanding the issue at hand. Everything else is an irrelevant noise that can hamper understanding. I would think we should not use such aliases in the examples.

index.html
```diff
@@ -8020,14 +8020,13 @@ <h3>Graph Containers</h3>
...
-          "isa": "@type",
-          ****"said": {"@id": "http://example.com/said", "@container": "@graph"}****
+          "said": "http://example.com/said",
+          ****"wrote": {"@id": "http://example.com/wrote", "@container": "@graph"}****
         },
-        "id": "Epimenides",
-        ****"said"****: {
-          "id": "Epimenides",
-          "isa": "Liar"
+        "@id": "Skakespeare",
```
dlongley: `Shakespeare` (here and elsewhere)

dlongley:
@pchampin -- good suggestion w/Shakespeare!

PR #145, opened 2019-03-19
describing property-based indexing as a variant of data indexing

pchampin:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/145.html" title="Last updated on Mar 26, 2019, 10:55 PM UTC (f2dc97d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/145/dbb3103...f2dc97d.html" title="Last updated on Mar 26, 2019, 10:55 PM UTC (f2dc97d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6580,14 +6580,14 @@ the data type to be specified explicitly with each piece of data.</p>
           "@context": {
             "@version": 1.1,
             "schema": "http://schema.org/",
-            "dc": "http://purl.org/dc/elements/1.1/",
+            "dc11": "http://purl.org/dc/elements/1.1/",
             "name": "schema:name",
             "body": "schema:articleBody",
             "words": "schema:wordCount",
             "post": {
               "@id": "schema:blogPost",
               ****"@container": "@index",
-              "@index": "dc:language"****
+              "@index": "dc11:language"****
             }
           },
           "@id": "http://example.com/",
@@ -6595,7 +6595,7 @@ the data type to be specified explicitly with each piece of data.</p>
           "name": "World Financial News",
           "post": {
             "en": {
-            ####↑ "en" will add `"dc:language": "en"` when expanded####
+            ####↑ "en" will add `"dc11:language": "en"` when expanded####
               "@id": "http://example.com/posts/1/en",
               "body": "World commodities were up today with heavy trading of crude oil...",
               "words": 1539
@@ -6687,7 +6687,7 @@ the data type to be specified explicitly with each piece of data.</p>
         </tr>
         <tr>
           <td>http://example.com/posts/1/de</td>
-          <td>dc:language</td>
+          <td>dc11:language</td>
           <td>de</td>
           <td></td>
         </tr>
@@ -6705,7 +6705,7 @@ the data type to be specified explicitly with each piece of data.</p>
         </tr>
         <tr>
           <td>http://example.com/posts/1/en</td>
-          <td>dc:language</td>
+          <td>dc11:language</td>
           <td>en</td>
           <td></td>
         </tr>
@@ -6719,7 +6719,7 @@ the data type to be specified explicitly with each piece of data.</p>
          data-to-rdf>
       <!--
         @prefix schema: <http://schema.org/> .
-        @prefix dc: <http://purl.org/dc/elements/1.1/language> .
+        @prefix dc11: <http://purl.org/dc/elements/1.1/language> .
         @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
         
         <http://example.com/> a schema:Blog;
@@ -6731,13 +6731,13 @@ the data type to be specified explicitly with each piece of data.</p>
         schema:articleBody
         "Die Werte an Warenbörsen stiegen im Sog eines starken Handels von Rohöl...";
         schema:wordCount 1204;
-        ****dc:language "de"****.
+        ****dc11:language "de"****.
         
         <http://example.com/posts/1/en>
         schema:articleBody
         "World commodities were up today with heavy trading of crude oil...";
         schema:wordCount 1539;
-        ****dc:language "en"****.
+        ****dc11:language "en"****.
         -->
     </pre>
   </aside>
--- a/index.html
+++ b/index.html
@@ -6741,6 +6741,11 @@ the data type to be specified explicitly with each piece of data.</p>
         -->
     </pre>
   </aside>
+
+  <p class="note">When using property-based data indexing, <a>index maps</a> can only be used
+    on <a>node objects</a>, not <a>value objects</a> or <a>graph objects</a>.
+    <a>Value objects</a> are restricted to have only certain keys and do not support
+    arbitrary properties.</p>
 </section>
 
 </section>
--- a/index.html
+++ b/index.html
@@ -11060,6 +11060,32 @@ the data type to be specified explicitly with each piece of data.</p>
   </section>
 
   <section class="changed">
+    <h2>Property-based Index Maps</h2>
+
+    <p>A property-based <a>index map</a> is a variant of <a>index map</a>
+      were indexes are semantically preserved in the graph as property values.
+      A property-based <a>index map</a> may be used as a term value within a <a>node object</a>
+      if the <a>term</a> is defined with <code>@container</code> set to <code>@index</code>,
+      or an array containing both <code>@index</code> and <code>@set</code>,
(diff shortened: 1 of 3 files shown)


gkellogg:
Looks like the Travis config needs a tweak; I’ll look into that.

index.html
```diff
@@ -6441,6 +6441,195 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
+      <button data-selects="statements">Statements</button>
+      <button data-selects="turtle">Turtle</button>
+      <a class="playground" target="_blank"></a>
+    </div>
+    <pre class="original selected nohighlight" data-transform="updateExample">
+      <!--
+        {
+          "@context": {
+            "schema": "http://schema.org/",
+            "dc": "http://purl.org/dc/elements/1.1/",
```
gkellogg: In the front of the spec, we describe our namespaces, and use "dc" for dcterms, not elements. It would be good to keep these in sync (there's enough confusion out there on terms vs. elements, and elements is generally considered archaic).

index.html
```diff
@@ -6441,6 +6441,195 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
+              ****"@container": "@index",
+              "@index": "dc:language"****
+            }
+          },
+          "@id": "http://example.com/",
+          "@type": "schema:Blog",
+          "name": "World Financial News",
+          "post": {
+            "en": {
+              "@id": "http://example.com/posts/1/en",
```
gkellogg: Probably worth adding a comment, such as the following

    ####↑ "en" will expand to `dc:language`####

gkellogg:
We'll also need to update the definition of `@index` to show that it can be used in this way, and the grammar section on term definitions.

index.html
pchampin: My understanding is that `dct:language` expects a resource, while `dce:language` expects a string. That's why I chose the later. But I agree that I should not reuse the `dc` prefix if it is used differently in the spec. I'll change it to `dce`.

pchampin:
> We'll also need to update the definition of `@index` to show that it can be used in this way, and the grammar section on term definitions.

Yes, but we should probably discuss before on a call whether we want to keep `@index` or introduce another key (you and @iherman seemed to have hesitation about reusing `@index` here.

gkellogg:
As I'm implementing this, I note some other things that seem to be requirements, or if not, add much more complication:

Consider the following example:

```json
{
  "@context": {
    "@version": 1.1,
    "@vocab": "http://example.com/",
    "author": {"@container": "@index", "@index": "prop"},
    "prop": {"@type": "@vocab"}
  },
  "@id": "http://example.com/article",
  "author": {
    "regular": {"@id": "http://example.org/person/1"},
    "guest": [
      {"@id": "http://example.org/person/2"},
      {"@id": "http://example.org/person/3"}
    ]
  }
}
```

* We actually need to use the term definition for the `@index` value, so that we know how to interpret string values of the map. So "regular" and "guest" are expanded as vocabulary-relative IRIs.
* You can't use a property to index a value object, so if in the above example, the values of "regular" and "guest" were't node objects (or strings which expand to node objects or graph objects), it would be an error.
* When compacting, we need to remember "prop" as a term, rather than it's expanded value; this way, we can find values of that term to extract for indexing.

Additionally, a node object containing both `@graph` and some other property is not a _graph object_, as defined in the spec. Therefore, indexing can't apply if `@container` includes both `@index` and `@graph` and `@index` is defined on the term. Logically, this might work, but it would wreak havoc on the already overly complicated compaction algorithm, where the notion of _graph object_ and _simple graph object_ come into play.

There are likely more complications, and possibly some corner cases I've neglected.

pchampin:
Gosh, I knew a bunch of worms were lurking from inside that shiny can!

Regarding the values in the index map, I'm ok to contrain them to be simple object nodes (no `@graph`, non `@value`).

Regarding the property used for indexing, my idea was that it would always have simple string values. It is tempting to use vocabulary-relative IRIs, but is there a real use case for that (beyond type indexing or id indexing)?

I think I would prefer to keep term definitions out of the picture for the indexing property, because they would bring too much complexity and too many corner cases (`@reverse`?). We could even require the value of `@index` to be an absolute IRI. This would be ugly and inconvenient, but it would hopefully prevent users from expecting a term definition to apply (although one could of course still try to add a term definition identified by the full IRI...).

> When compacting, we need to remember "prop" as a term, rather than it's expanded value; this way, we can find values of that term to extract for indexing.

Sorry, I don't get it. In the expanded form, we only have expanded values, right?

gkellogg:
I think keeping them as term definitions works. The important thing is that, in the compact view, the value of the property is expressed as a string.

The point on prop being a term when compacting is so that we know how to compact the value of "prop" to use as an index. Of course, if we only allowed simple values we could avoid this, but it seems to me to be useful to allow anything that compacts as a string to be used as an index. The main complicating area is for named graph indexing, but I think this is a vanishingly small use case.

I did consider that as an alternative to leveraging `"@container": "@index"`, and using an `"@index"` value, we could allow `"@container": "prop"`, where "prop" is a term or expands to an IRI. There's not a lot of shared logic in the algorithms for the property-index vs. `@index` use cases.

index.html
```diff
@@ -6553,6 +6553,195 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
```
```suggestion
              "@id": "http://example.com/posts/1/en",
              ####↑ "en" will expand to `dc:language`####
```

index.html
```diff
@@ -6553,6 +6553,195 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
+      However, in some situations,
+      the keys used to index objects are semantically linked to these objects,
+      and should be preserved not only syntactically, but also semantically.
+    </p>
+    <p>If the <a>processing mode</a> is set to <code>json-ld-1.1</code>,
+      <code>"@container": "@index"</code> in a term description can be accompanied with
+      an <code>"@index"</code> key. The value of that key must map to an <a>IRI</a>,
+      which identifies the semantic property linking each object to its key.
+    </p>
+    <aside class="example ds-selector-tabs"
```
gkellogg: We should add some description of the restrictions on using property-based indexing, particularly on value objects and the inability to use on graph maps.

index.html
```diff
@@ -6552,6 +6552,196 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
+              "@id": "schema:blogPost",
+              ****"@container": "@index",
+              "@index": "dc:language"****
+            }
+          },
+          "@id": "http://example.com/",
+          "@type": "schema:Blog",
+          "name": "World Financial News",
+          "post": {
+            "en": {
```
```suggestion
            "en": {
            ####↑ "en" will add `"dc:language": "en"` when expanded####
```

index.html
gkellogg: Fixed in 2ed383d58b4b00589be1e92838c62876e78c5ee6.

index.html
```diff
@@ -11119,9 +11119,10 @@ <h2>Type Maps</h2>
       or an array containing both <code>@type</code> and <code>@set</code>.
       The keys of a <a>type map</a> MUST be <a>IRIs</a>
       (<a>relative IRI</a>, <a>compact IRI</a> (including <a>blank node identifiers</a>), or <a>absolute IRI</a>),
-      the <a>keyword</a> <code>@none</code>,
-      or a <a>term</a> which expands to <code>@none</code>,
-      and the values MUST be <a>node objects</a>.</p>
+      <a>terms</a>,
```
pchampin: the previous version limited this to "terms expaning to @none", but for type-map, it seems to me that any term can be used (at least, this works in the playground and the Ruby distiller)

index.html
```diff
@@ -11122,7 +11122,7 @@ <h2>Type Maps</h2>
       <a>terms</a>,
       or the <a>keyword</a> <code>@none</code>,
       and the values MUST be <a>node objects</a>
-      or <a>strings</a> which expand to node obects.</p>
+      or <a>strings</a> which expand to <a>node objects</a>.</p>
```
pchampin: typo apart, I added this line because type maps may have plain IRIs as values.
Not entirely sure about the phrasing though.

index.html
gkellogg: This line is in the Type Maps section, which has always allowed terms.

gkellogg:
Okay, I think that this, and w3c/json-ld-api#78 are ready. We can always tweak further.

PR #152, opened 2019-03-27
Better phrasing of object equality

pchampin:
I think the current phrasing is actually inaccurate. It seemed to imply (at least to me, a non-native speaker) that *one* equal member was enough for the objects to be considered equal.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/152.html" title="Last updated on Mar 27, 2019, 4:54 PM UTC (36ece32)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/152/1e85d7a...36ece32.html" title="Last updated on Mar 27, 2019, 4:54 PM UTC (36ece32)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -11409,8 +11409,12 @@ the data type to be specified explicitly with each piece of data.</p>
           <li>If <var>A</var> and <var>B</var> are both <a>objects</a>,
             both <var>A</var> and <var>B</var> have the same number of <a>members</a>,
             and each <a>member</a> in <var>A</var> is equal to a corresponding <a>member</a> in <var>B</var>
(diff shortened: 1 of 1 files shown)


iherman:
I approve @pchampin's change, but re-reading the whole thing I wonder whether there isn't something missing (from a spec point of view). It says "keys are identical". Keys are strings, but we may still want to refer to the ECMAScript document as for what it means that they are identical. The reason that the equality of strings is not that obvious in a Unicode world, but the ECMAScript does the right thing talking about the code points...

iherman:
Thanks @pchampin !

Issue #136, opened 2019-02-21
Additional restriction to `@sealed` term clearing


dlongley:
We've completed a preliminary implementation of `@sealed` terms here:

https://github.com/digitalbazaar/jsonld.js/pull/289

Our implementation:

* Allows terms to be sealed using `"@sealed": true` in a term definition (or not sealed with `"@sealed": false`)
* Allows setting the default `"@sealed"` for all terms in an `@context` via `"@sealed": true` at the top level
* Does not affect existing term clearing rules when there are no `@sealed` terms
* **Allows clearing `@sealed` terms ONLY via a new property term definition with a scoped `@context` (note: and the new term does not need to be sealed)**

Note that the last item is a slight tweak (additional restriction) on what is in the spec. We believe this approach better matches the use cases and expectations of the target audience for JSON-only application developers. The high-level idea is that a `@sealed` term is assured to maintain its meaning unless cleared by a new term (with a scoped context) that creates a new branch in the JSON tree.

This approach:

* Provides sufficient assurance for the meaning of `@sealed` terms without a need for JSON-LD processing: a JSON-only dev can read a spec and write **safe** code that traverses JSON-LD data via `@sealed` terms ... without a need for JSON-LD processing, instead using only validation rules (e.g. json-schema)
* Allows extensions to the JSON tree to occur **safely** in an idiomatic way where `@sealed` terms can only be cleared under new branches in the JSON tree (matches common JSON developer intuitions/expectations)
* Very closely mirrors what @pchampin articulated so well a while back but has the added benefit of being simpler to implement: there is no need to track the `@context` from whence `@sealed` terms came

We'd like to see the spec updated to restrict clearing `@sealed` terms in the way articulated here and implemented in the above link. This approach seems to fully address the original need for the feature and has a simple implementation.

gkellogg:
This looks like a middle ground between keeping track of sealed terms from different contexts and providing some impediment to simply wiping out sealed terms, which only takes effect when you've stepped into a different part of the JSON tree.

iherman:
@dlongley, thanks. Just for my understanding, can you propose alternate versions of the examples 42-44? I am. e.g., not sure what you mean by "clearing".

[...]

PR #156, opened 2019-04-15
reflected the decisions about #136

pchampin:
NB: marked example 44 as ignored, because it is not supported by the current implementation

### DIFF
--- a/index.html
+++ b/index.html
@@ -3325,7 +3325,7 @@ specified. The full <a>IRI</a> for
     JSON-LD 1.1 allows term definitions to be <em>protected</em>.
     </p>
   <p>A <dfn>protected term definition</dfn> is a term definition with a member <code>@protected</code> set to <code>true</code>.
-    It generally prevents further contexts to override this term definition,
+    It generally prevents further contexts from overriding this term definition,
     either through a new definition of the same term,
     or through clearing the context with <code>"@context": null</code>.
     Such attempts will raise an error and abort the processing
--- a/index.html
+++ b/index.html
@@ -3334,7 +3334,7 @@ specified. The full <a>IRI</a> for
   </p>
 
 
-  <pre class="example" data-transform="updateExample"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3325,89 +3325,47 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
     JSON-LD 1.1 allows term definitions to be <em>protected</em>.
     </p>
   <p>A <dfn>protected term definition</dfn> is a term definition with a member <code>@protected</code> set to <code>true</code>.
-    It prevents further contexts to override this term definition.
-    </p>
+    It generally prevents further contexts to override this term definition,
```
```suggestion
    It generally prevents further contexts from overriding this term definition,
```

gkellogg:
Looks like PR Preview isn't working again :(.

index.html
```diff
@@ -3325,89 +3325,47 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
-    <div class="selectors">
-        <button class="selected" data-selects="original">Original</button>
-        <button data-selects="expanded">Expanded</button>
-        <button data-selects="statements">Statements</button>
-        <button data-selects="turtle">Turtle</button>
-        <a class="playground" target="_blank"></a>
-    </div>
-    <pre class="original selected nohighlight" data-transform="updateExample">
-      <!--
+  <pre class="example" data-transform="updateExample"
```
gkellogg: [According to ReSpec](https://github.com/w3c/respec/wiki/example) we can use `illegal-example` for examples of things that should not evaluate; perhaps this is a good place to use it.

index.html
```diff
@@ -3325,89 +3325,47 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
```
```suggestion
  <pre class="illegal-example example" data-transform="updateExample"
```

```suggestion
  <pre class="illegal-example" data-transform="updateExample"
```

PR #157, opened 2019-04-15
a number of small editorial changes

pchampin:
none of them should be too controversial, but I'd rather have a review, just in case...

### DIFF
--- a/index.html
+++ b/index.html
@@ -394,7 +394,7 @@
    <dd>A JSON-LD document is always a valid JSON document. This ensures that
     all of the standard JSON libraries work seamlessly with JSON-LD documents.</dd>
    <dt>Expressiveness</dt>
-   <dd>The syntax serializes directed graphs. This ensures that almost
+   <dd>The syntax serializes labeled directed graphs. This ensures that almost
     every real world data model can be expressed.</dd>
    <dt>Terseness</dt>
    <dd>The JSON-LD syntax is very terse and human readable, requiring as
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1578,7 +1578,7 @@ <h1>Advanced Concepts</h1>
     describe such value typing.</p></dd>
   <dt>Value ordering</dt>
   <dd><p>In JSON, a property with an array value implies an implicit order;
-    arrays in JSON-LD do not provide an ordering of the contained elements by
+    arrays in JSON-LD do not convey any ordering of the contained elements by
```
pchampin: It might just be me, being not a native speaker,
but as I see it, an array always *provide* an ordering.
It is just that we decide to to ignore it (unless they appear as the value of `@list`).

gkellogg: I think both wordings are fairly close, and it comes down to editor's choice.

gkellogg:
For this to build cleanly, changes need to be synced to json-ld-wg/common. And, of course, eventually need to be deployed to other repos. I'll make the change I suggested above and do just that.

index.html
```diff
@@ -4241,12 +4241,12 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
 
   ex1:barney ex2:mnemonic "the sidekick" .
 
-  [ ex2:fred ex1:barney, ex1:barney] .
+  [] ex2:fred ex1:barney, ex1:barney.
```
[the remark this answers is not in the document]
pchampin: In general I would prefer the original formulation in a turtle file.

But in the text, we quote a turtle snippet, calling it "a single triple". I thought that `[] ex:2 fred ex1:barney` made the 3 parts of the triple more obvious  than `[ ex2:fred ex1:barney1 ]`.

For the sake of consistency, I changed the Turtle example accordingly.

PR #161, opened 2019-04-16
Add caveat on numbers being represented as double

gkellogg:
if they're too large to represent as integers.

### DIFF
--- a/index.html
+++ b/index.html
@@ -10291,7 +10291,7 @@ the data type to be specified explicitly with each piece of data.</p>
       the scope of a <a>JSON-LD document</a>. Blank node identifiers begin with
       <code>_:</code>.</li>
     <li>A <a>JSON-LD value</a> is a <a>typed value</a>, a <a>string</a> (which is interpreted
-      as <a>typed value</a> with type <code>xsd:string</code>), a <a>number</a>
+      as a <a>typed value</a> with type <code>xsd:string</code>), a <a>number</a>
       (<a>numbers</a> with a non-zero fractional part, i.e., the result of a modulo&#8209;1 operation,
       <span class="changed">or which are too large to represent as integers,</span>
       are interpreted as <a>typed values</a> with type <code>xsd:double</code>, all other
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -10293,6 +10293,7 @@ <h1>Data Model</h1>
     <li>A <a>JSON-LD value</a> is a <a>typed value</a>, a <a>string</a> (which is interpreted
       as <a>typed value</a> with type <code>xsd:string</code>), a <a>number</a>
```
```suggestion
      as a <a>typed value</a> with type <code>xsd:string</code>), a <a>number</a>
```

index.html
```diff
@@ -10291,8 +10291,9 @@ <h1>Data Model</h1>
       the scope of a <a>JSON-LD document</a>. Blank node identifiers begin with
       <code>_:</code>.</li>
     <li>A <a>JSON-LD value</a> is a <a>typed value</a>, a <a>string</a> (which is interpreted
-      as <a>typed value</a> with type <code>xsd:string</code>), a <a>number</a>
+      as a <a>typed value</a> with type <code>xsd:string</code>), a <a>number</a>
       (<a>numbers</a> with a non-zero fractional part, i.e., the result of a modulo&#8209;1 operation,
+      <span class="changed">or which are too large to represent as integers,</span>
```
rubensworks: I think it is important to specify what kind of integer this is (XSD, JSON, programming language integer, ...).
It looks like the XSD spec defines [the value space of `xsd:integer` as an infinite set](https://www.w3.org/TR/xmlschema-2/#integer), which means that there is no maximum (or minimum) value.
Next to that, the [JSON RFC also does not place a strict limit on integer/number sizes](https://tools.ietf.org/html/rfc7159#section-6), but merely allows limits to be set by software.

Unless this has been defined somewhere already, I think it would be good to define integer sizes in the JSON-LD spec. To be consistent with most programming languages, values for 32-bit integer sizes may be useful. Perhaps this can be a soft limit, to be in line with JSON.

Side-note: Next to mentioning values that are _too large_, values that are _too small_ should also be considered.

iherman: Are there such limit in the RDF model? I do not remember having one, in which case I do not think that, formally, JSON-LD could impose any limit...

gkellogg:
This corresponds to https://github.com/w3c/json-ld-api/pull/79, and is a practical consideration of round-tripping issues. In theory, JSON numbers can be of arbitrary length, but in practice, implementations use IEEE floating point representations for numbers (JS in particular, but also Ruby). Consequently, there is a limit to the size of an integer that JSON can represent with fidelity to be < `1e21`.

However, it may be that we don't need to reflect such interoperability issues in the Syntax document.

index.html
gkellogg: Text was added to refer to Data Round Tripping for limitations on integer representations.

PR #168, opened 2019-04-29
ack script data and generated file

iherman:
@gkellogg @BigBlueHat @azaroth42 

I have generalized the acknowledgment generation script to configure it and run it more generally (it was, so far, tailored at the Web Publication document). The script itself is in its [own repo now](https://github.com/w3c/publ_ack). It is not yet final (and no documentation yet), but the essence is there.

- There is a generic [configuration file](https://github.com/w3c/json-ld-wg/blob/master/assets/publ_ack_config.json). It currently has references to the files that I have created for the syntax document (and using the final URL-s as they will be, eventually). This will have to be extended to the framing and the api documents.
- There are two files that are created for each document:
  - the explicit list in `separate_acks.json` which lists those that are to be mentioned as really authors and contributors. I have created the file in this PR, listing all those who have been at least as one of our calls this year.
  - a template file in `ack_pattern.html` that contains the slots for the two lists: the ones called out explicitly, and all the others. This will be documented, but it is fairly self-evident.
- I have generated the file for this document in `acknowledgment.html`. This can then be included into the final document via respec, using:

```
<section id="ack" data-include="common/ack-script/acknowledgements.html" data-include-replace="true"></section>
```

I am sure there are refinements to be done on the script, but the essence it there. I wanted to share it asap to get the feedbacks in time.

If you install this and want to run locally, two things to remember:

- You will have to create a `~/.publ_ack.json` which should include, at the minimum, the `api_key` for accessing the W3C API-s that gives one access to these data (see https://w3c.github.io/w3c-api/).
- If you run it, be patient. The script has to issue a sh...load of API calls to get everything; the API is not very optimal. I tried to use parallel promises wherever I could (and it does speed up a lot), but it takes its time...

---

I did not want to merge this 'just like that', though it does not affect any of the publication files...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/168.html" title="Last updated on May 1, 2019, 5:04 PM UTC (855c2c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/168/449c65f...855c2c5.html" title="Last updated on May 1, 2019, 5:04 PM UTC (855c2c5)">Diff</a>


gkellogg:
Thanks Ivan, this looks promising. Note that files in the common directory are expected to be replicated (by hand) to all of the different repositories. If these files are "common", in the sense that they are used _for_ all the repositories, but not expected to reside _within_ those repositories, maybe we should use a different directory.

I note that Rob Sanderson, David Longley and David Lehn appear in both sections, somewhat differently, and without their organizational attribution; presumably these should be the same.

The wording indicates that the members of the WG section would exclude those in the significant contributors section (because of the "Additionally"), and probably shouldn't include the editors.

It also might be nice to flag Rob and Benjamin as chairs, as you've got that info in the JSON config file.

iherman:
@gkellogg, I will look at the minor issues there. I suspect the reason Rob and others to appear twice is that their name in the official database is different than what I used in the curated file.

As for the place in the repo, please change it as you prefer. They are indeed not common; they can differ from one document to the other. Eg, PA would have to appear in the manually created list in the frame document because he is not an editor of that one.

BigBlueHat:
Looks great overall, @iherman!

This probably won't shock anyone that I'm wishing we could somehow overlap the `nicknames.json` and the new `separate_acks.json` file. Maybe there's cause for a new WG-level `people.json` file in JSON-LD that could more completely (as needed) contain these various relationships with the various docs.

If the answer is "nah, that's too much work." We should (re)consider the value claims we use when promoting JSON-LD...and/or improve the tooling...

I'm happy to help craft such a thing (if it's wanted), fwiw. 😃

iherman:
@BigBlueHat 

> This probably won't shock anyone that I'm wishing we could somehow overlap the nicknames.json and the new separate_acks.json file. Maybe there's cause for a new WG-level people.json file in JSON-LD that could more completely (as needed) contain these various relationships with the various docs.

This may become a bit complex. The `nicknames.json` file is to have and maintain per group. The `separate_acks.json` file is per document. It *can* of course be done, but it may lead to some refactorization of several things with a slightly more complex JSON structure.

Such a grand unification would make sense, of course, but it is quite a lot of work and should be done in a larger framework of tools to handle WG tasks which goes beyond what I should do (there are dedicated people for that at W3C...).

iherman:
@gkellogg I have 

- made some changes (and handled a bug) in the script, the result has the chairs properly flagged
- the ack file had to be improved; the names in the list _must_ be identical to the name used in the W3C user database (and some of those were slightly different, like 'Robert Sanderson' as opposed to 'Rob Sanderson'...)
- I have added an extra possible flag for 'editors' in the ack file; those names are removed from the global listing, but not displayed in the first, explicit list either

I have updated the results. I have not changed the place in the file hierarchy, I let you decide what you would prefer to have. I can then generate the relevant files for the API and the Frames' documents, too.

iherman:
One question: there was a feature in the script that I documented out, namely that one could get all the WG members in the listing, _including_ the former WG members. I am not sure that is necessary. WDYT?

gkellogg:
I think it's reasonable that acknowledgements are limited to this WG and the CG without invoking all of the acknowledgements from the 1.0 WG, which of course, live on in that document.

I've integrated it into the ReSpec doc for people to inspect in place.

iherman:
> 
> I've integrated it into the ReSpec doc for people to inspect in place.

Once this is accepted and merged, I will add the configuration files to the other repositories. I guess it will be all right to 'just' commit the changes without going through the PR phase, right?

gkellogg:
> > I've integrated it into the ReSpec doc for people to inspect in place.
> 
> Once this is accepted and merged, I will add the configuration files to the other repositories. I guess it will be all right to 'just' commit the changes without going through the PR phase, right?

Fine with me to do simple commit to master.

PR #195, opened 2019-06-19
Describe that type-scoped contexts are only in affect on the node objects in which they're used

### LINKED ISSUE #174: Unexpected @type scope behavior (term definitions persist throughout JSON tree)
I was trying to use type-scoped contexts to define a `@context` and was surprised to discover that any type-scoped terms that get defined in the active context continue to be defined beyond the object with the matching `@type`. I think this is very unexpected behavior from an OOP modeling perspective. Also, it is very problematic for `@protected` terms, as it means that you can't model objects of one type that contain objects of another type when there is a commonly used JSON key (that may or may not have the same term definition) when terms are protected.

A playground example:

```js
{
  "@context": {
    "@version": 1.1,
    "@vocab": "ex:",
    "@protected": true,
    "Library": {
      "@context": {
        "book": "library:book",
        "name": "library:name"
      }
    },
    "Person": {
      "@context": {
        "name": "person:name"
      }
    }
  },
  "@id": "the:library",
  "@type": "Library",
  "book": {
    "@id": "the:book",
    "about": {
      "@id": "the:person",
      "@type": "Person",
      "name": "Oliver Twist",
      "book": "unexpectedly defined as library:book!"
    }
  }
}
```

Produces these quads:

```
<the:book> <ex:about> <the:person> .
<the:library> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <ex:Library> .
<the:library> <library:book> <the:book> .
<the:person> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <ex:Person> .
<the:person> <library:book> "unexpectedly defined as library:book!" .
<the:person> <person:name> "Oliver Twist" .
```

http://tinyurl.com/y2x4szzb

If you use `@protected` here, you get an error (which I also find unexpected):

```js
{
  "@context": {
    "@version": 1.1,
    "@vocab": "ex:",
    "@protected": true,
    "Library": {
      "@context": {
        "@protected": true,
        "book": "library:book",
        "name": "library:name"
      }
    },
    "Person": {
      "@context": {
        "@protected": true,
        "name": "person:name"
      }
    }
  },
  "@id": "the:library",
  "@type": "Library",
  "book": {
    "@id": "the:book",
    "about": {
      "@id": "the:person",
      "@type": "Person",
      "name": "Oliver Twist",
      "book": "unexpectedly defined as library:book!"
    }
  }
}
```

That error happens *even if* `name` is defined the same way for both types.

I suspect that type-scoped terms behave this way because it was easy to implement, but I think it is very surprising behavior that may not have been exposed yet due to limited examples.

It's possible that there's an easy fix for this. I think we should change this behavior so that we track whether a term definition in the active context was defined via a type-scoped context and whether or not it replaced a non-type-scoped term when it did so. Then, whenever traversing into one of the typed object's properties during processing, we revert all type-scoped terms to their previous definitions which may mean setting them to `null` (clearing them) if they were previously undefined. Then processing can continue as normal.

gkellogg:
Add changes for type-scoped contexts.

Fixes #174.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/195.html" title="Last updated on Jun 19, 2019, 11:14 PM UTC (08b3a1c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/195/16c60eb...08b3a1c.html" title="Last updated on Jun 19, 2019, 11:14 PM UTC (08b3a1c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3410,6 +3410,9 @@ In this example, the <a>compact IRI</a> form is used in two different ways.
     the type is used; the previous in-scope <a>contexts</a> are placed back into
     effect when traversing into another <a>node object</a>.</p>
 
+  <p class="note">Any property-scoped or local contexts that were introduced in the <a>node object</a>
+    would still be in effect when traversing into another <a>node object</a>.</p>
+
   <p>When expanding, each value of <code>@type</code> is considered
     (ordering them lexicographically) where that value is also a <a>term</a> in
     the <a>active context</a> having its own <a>scoped context</a>. If so, that
--- a/index.html
+++ b/index.html
@@ -3405,7 +3405,10 @@ In this example, the <a>compact IRI</a> form is used in two different ways.
     relate things of different types, where the vocabularies in use within
     different entities calls for different context scoping. For example,
     <code>hasPart</code>/<code>partOf</code> may be common terms used in a document, but mean
-    different things depending on the context.</p>
+    different things depending on the context.
+    A <a>context</a> scoped on <code>@type</code> is only in effect for the <a>node object</a> on which
+    the type is used; the previous in-scope <a>contexts</a> are placed back into
+    effect when traversing into another <a>node object</a>.</p>
 
   <p>When expanding, each value of <code>@type</code> is considered
     (ordering them lexicographically) where that value is also a <a>term</a> in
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3405,7 +3405,10 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
     relate things of different types, where the vocabularies in use within
     different entities calls for different context scoping. For example,
     <code>hasPart</code>/<code>partOf</code> may be common terms used in a document, but mean
-    different things depending on the context.</p>
+    different things depending on the context.
+    A <a>context</a> scoped on <code>@type</code> is only in effect for the <a>node object</a> on which
+    the type is used; the previous in-scope <a>contexts</a> are placed back into
+    effect when traversing into another <a>node object</a>.</p>
```
pchampin: To be precise, we should add something like "amended by the appropriate property-scoped context and/or local context, if any." But I'm affraid that might confuse the reader more than anything else...

dlongley: Perhaps: `Note that any property-scoped or local contexts that were introduced in the node object would still be in effect when traversing into another node object.`

gkellogg: Applied @dlongley's suggestion.

PR #197, opened 2019-06-21
Remove index.html.orig.

davidlehn:
@BigBlueHat Looks like the `index.html.orig` slipped in a commit by mistake:
https://github.com/w3c/json-ld-syntax/commit/5acab54b1868a68397dcfaf043fad4ea9bdc49c6


BigBlueHat:
Gah! Didn't even realize that was still hanging around. 😝 Good catch!

PR #200, opened 2019-07-11
Add section for `@propagate` and stub for `@import`.

### LINKED ISSUE #174: Unexpected @type scope behavior (term definitions persist throughout JSON tree)

gkellogg:
Fixes #174.

(Still requires description of `@source`. Examples will fail until changes made to Ruby implementation, pending agreement on this and w3c/api#112.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/200.html" title="Last updated on Jul 12, 2019, 3:58 PM UTC (0384304)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/200/f5741e4...0384304.html" title="Last updated on Jul 12, 2019, 3:58 PM UTC (0384304)">Diff</a>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,10 @@
 source "https://rubygems.org"
 
 gem 'json-ld', github: 'ruby-rdf/json-ld', branch: 'develop'
+gem 'json-ld-preloaded'
 gem 'nokogiri'
 gem 'nokogumbo'
 gem 'linkeddata'
 gem 'colorize'
 gem 'rake'
+gem 'byebug'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: git://github.com/ruby-rdf/json-ld.git
-  revision: 74a05a9f623c5a9e02cbe42f053f9d09181fee97
+  revision: 813e1ff914cc1a1025db14de392fd738b4cbb674
   branch: develop
   specs:
     json-ld (3.0.2)
@@ -18,6 +18,7 @@ GEM
     bcp47 (0.3.3)
       i18n
     builder (3.2.3)
+    byebug (11.0.1)
     colorize (0.8.1)
     concurrent-ruby (1.1.5)
     connection_pool (2.2.2)
@@ -26,7 +27,7 @@ GEM
       sxp (~> 1.0)
     equivalent-xml (0.6.0)
       nokogiri (>= 1.4.3)
-    haml (5.0.4)
+    haml (5.1.1)
       temple (>= 0.8.0)
       tilt
     hamster (3.0.0)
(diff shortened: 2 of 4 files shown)


gkellogg:
Needs some work on examples and to update `@source` to `@import`, which I'll take care of presently.

PR #201, opened 2019-07-12
Add sourced context section.

dlongley:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/201.html" title="Last updated on Jul 12, 2019, 3:56 PM UTC (804c257)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/201/39ca24f...804c257.html" title="Last updated on Jul 12, 2019, 3:56 PM UTC (804c257)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3747,7 +3747,7 @@ In this example, the <a>compact IRI</a> form is used in two different ways.
   </pre>
 
   <p>Again, the effect would be the same as if the entire sourced <a>context</a>
-    had been copied into the type-scoped <a>context</a>:</p>
+    had been copied into the <a>context</a>:</p>
 
   <pre class="example nohighlight" data-transform="updateExample"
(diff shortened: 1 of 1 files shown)


dlongley:
I don't think that last normative bit about requiring a sourced context to be an object should be in a note ... but not sure where it goes (I was copying another section of spec text). It also has a typo in it that can be fixed when we address that.

dlongley:
Ha! Ok.

gkellogg:
> I don't think that last normative bit about requiring a sourced context to be an object should be in a note ... but not sure where it goes (I was copying another section of spec text). It also has a typo in it that can be fixed when we address that.

Yes, it probably should be outside the note, but that section is informative, in any case. The only place to make a normative statement would be in the grammar, which is a little more challenging.

PR #205, opened 2019-07-18
Example of using xref

marcoscaceres:
(don't merge this :) ... just examples)

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -24,7 +24,7 @@
         <dfn data-cite="INFRA#ordered-map" class="preserve">map</dfn> (see [[INFRA]]),
         composed of <dfn data-cite="INFRA#map-entry" data-lt="map entry|entry" data-lt-noDefault class="preserve">entries</dfn> with key/value pairs.</p>
       <p>In the <a data-cite="JSON-LD11-API#the-application-programming-interface">Application Programming Interface</a>,
-        a <a>map</a> is described using a [[WEBIDL]] <a data-cite="WEBIDL#dfn-dictionary">dictionary</a>.</p></dd>
+        a <a>map</a> is described using a [[WEBIDL]] <a>dictionary</a>.</p></dd>
     <dt class="changed">JSON-LD <a>internal representation</a></dt>
     <dd class="changed">
       The JSON-LD internal representation
--- a/common/terms.html
+++ b/common/terms.html
@@ -1,343 +1,348 @@
 <dl class="termlist" data-sort>
-  <dt><dfn data-cite="INFRA#list" class="preserve">array</dfn></dt><dd>
-    In the JSON serialization,
-    an <a>array</a> structure is represented as square brackets surrounding zero or more values.
-    Values are separated by commas.
-    In the <a>internal representation</a>,
-    a <a data-ld="array">list</a> (also called an <a>array</a>) is an <em>ordered</em> collection of zero or more values.
-    While JSON-LD uses the same array representation as JSON,
-    the collection is <em>unordered</em> by default.
-    While order is preserved in regular JSON arrays,
-    it is not in regular JSON-LD arrays unless specifically defined
-    (see <a data-cite="JSON-LD11#sets-and-lists">Sets and Lists</a>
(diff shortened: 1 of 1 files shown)


common/terms.html
```diff
@@ -1,343 +1,348 @@
...
+      A single colon comes after each name,
+      separating the name from the value.
+      A single comma separates a value from a following name.
+      In JSON-LD the names in an object MUST be unique.
+      <p>In the <a>internal representation</a> a <a>JSON object</a> is described as a
+        <dfn data-cite="INFRA#ordered-map" class="preserve">map</dfn> (see [[INFRA]]),
+        composed of <dfn data-cite="INFRA#map-entry" data-lt="map entry|entry" data-lt-noDefault class="preserve">entries</dfn> with key/value pairs.</p>
+      <p>In the <a data-cite="JSON-LD11-API#the-application-programming-interface">Application Programming Interface</a>,
+        a <a>map</a> is described using a [[WEBIDL]] <a>dictionary</a>.</p></dd>
+    <dt class="changed">JSON-LD <a>internal representation</a></dt>
```
marcoscaceres: Here is an example of xref auto-linking a common term.

(I think this document has Window's line endings, which Visual Studio Code removed)

common/terms.html
```diff
@@ -1,343 +1,348 @@
...
+      A name is a <a>string</a>.
+      A single colon comes after each name,
+      separating the name from the value.
+      A single comma separates a value from a following name.
+      In JSON-LD the names in an object MUST be unique.
+      <p>In the <a>internal representation</a> a <a>JSON object</a> is described as a
+        <dfn data-cite="INFRA#ordered-map" class="preserve">map</dfn> (see [[INFRA]]),
+        composed of <dfn data-cite="INFRA#map-entry" data-lt="map entry|entry" data-lt-noDefault class="preserve">entries</dfn> with key/value pairs.</p>
+      <p>In the <a data-cite="JSON-LD11-API#the-application-programming-interface">Application Programming Interface</a>,
+        a <a>map</a> is described using a [[WEBIDL]] <a>dictionary</a>.</p></dd>
```
marcoscaceres: Note here, just `<a>dictionary</a>`... don't need data-cite.

common/terms.html
```diff
@@ -1,343 +1,348 @@
...
+    <dt class="changed">JSON-LD <a>internal representation</a></dt>
+    <dd class="changed">
+      The JSON-LD internal representation
+      is the result of transforming a JSON syntactic structure
+      into the core data structures suitable for direct processing:
+      <a>arrays</a>, <a>maps</a>, <a>strings</a>, <a>numbers</a>, <a>booleans</a>, and <a>null</a>.</dd>
+    <dt>
+      <a>null</a>
```
marcoscaceres: Similarly here... just `<a>null</a>`.

common/terms.html
```diff
@@ -1,343 +1,348 @@
...
-    which is a <a>string</a>,
-    and a type,
-    which is an <a>IRI</a>.</dd>
-  <dt><dfn>value object</dfn></dt><dd>
-    A <a>value object</a> is a <a>map</a> that has an <code>@value</code> <a>entry</a>.</dd>
-  <dt><dfn class="preserve">vocabulary mapping</dfn></dt><dd>
-    The vocabulary mapping is set in the <a>context</a> using the <code>@vocab</code> key
-    whose value MUST be an <a>IRI</a> or <code>null</code>.</dd>
-</dl>
+    <dt><dfn data-cite="INFRA#list" class="preserve">array</dfn></dt><dd>
```
marcoscaceres: Note that this might confuse people... array and lists are different things.

gkellogg: Yes, but it was not feasible to change existing uses of list, which are associated with RDF lists and oradered arrays in JSON-LD. Array is also more compatible with general JSON and WebIDL nomenclature.

common/terms.html
gkellogg: It doesn't work for me, I get "Couldn't match "null" to anything in the document or to any other spec."

Issue #19, opened 2018-06-30
Indexing without a predicate


gkellogg:
For consideration by the JSON-LD 1.1 WG...

Assuming a nested set of resources where leaf nodes are frequently repeated, it is difficult to find the definition of the node after compaction.  Imagine a classification that is used on the second item in a list, and again on the 26th.  It would be nice to have a place to look up the label for the classification, instead of repeating it on both 2 and 26.  Similarly, information about repeated people, services, or anything else could benefit from this pattern.

As prior art, and use case for inclusion, [JSON API](http://jsonapi.org/format/#document-top-level) has the notion of "included" -- a slot where you can put resources that are included in others, such that developers can always know where to find them.  In my work, this has come up with repeated services in IIIF, and classifications, people and places when describing the cultural heritage objects they relate to.

The identifier map pattern is already in this space, but insufficient as it requires a predicate to map to, and the relationship is to a resource somewhere nested in the data structure, not the top level resource. There would also need to be framing support as an extension to `@embed:@never` such that the inclusions were not embedded in the object data, but with a pointer to where they should go.

Example data:

```
{
  "id": "1",
  "type": "eg:Thing-with-Items",
  "eg:items": [
    {
      "id":"2",
      "classification": "enum:c6",
      "service": "enum:s2",
    },
    { "id": "3...26 go here", "type": "eg:X"}, 
    {
      "id": "27",
      "classification": "enum:c6"
    }    
  ],
  "included": {
    "enum:c6": {"type": "eg:Type", "label": "Classification 6"},
    "enum:p1": {"type": "eg:Person", "label": "Person 1"},
    "enum:s2": {"type": "eg:Service", "label": "Login Service"}
  }  
}
```

Playground example with identifier map:  http://tinyurl.com/yd5z87xg

The inclusion term could either be a new keyword like `@id` (`@included` or `@inclusions`) that was then re-aliased in the context (to, e.g. `included`), or it could be a new keyword value for `@container` (`included: {@container: @included}`).  I think the former is the (slightly) better design, as it makes it more obviously a field rather than a data structure. It would only be usable in a resource that is not nested within other resources (e.g. the top level JSON object ... which might be in an array or @graph). Framing could then use the same keyword: `@embed: @included`.

Original issue: [Indexing without a predicate #650](https://github.com/json-ld/json-ld.org/issues/650)

gkellogg:
[...]

Issue #128, opened 2019-02-06
TriG graphs in JSON-LD


iherman:
I just try to test my understanding...

Consider the following TriG dataset:

```
@prefix a <http://example.org/a> .
@prefix b <http://example.org/b> .
@prefix c <http://example.org/c> .

<http://example.org/gr1> {
    a:something a:predicate "value1",
                b:predicate c:something1 .
}
<http://example.org/gr2> {
    a:something a:predicate "value2",
                b:predicate c:something2 .
}
```

The only way to express this in JSON-LD to repeat a `@context`, i.e.,

```json
[
    {
        "@context" : {
            "a": "http://example.org/a",
            "b": "http://example.org/b",
            "c":  "http://example.org/c",
            "b:predicate" : { "@type" : "@id" }
        },
        "@id" : "http://example.org/gr1",
        "@graph" : {
            "@id" : "a:something",
            "a:predicate" : "value1",
            "b:predicate" : "c:something1"
        }
    },
    {
        "@context" : {
            "a": "http://example.org/a",
            "b": "http://example.org/b",
            "c":  "http://example.org/c",
            "b:predicate" : { "@type" : "@id" }
        },
        "@id" : "http://example.org/gr2",
        "@graph" : {
            "@id" : "a:something",
            "a:predicate" : "value2",
            "b:predicate" : "c:something2"
        }
    }
]
```

I.e., the context must be repeated.; there is no way of using a "global" context. I am not sure whether this is a problem in practice, but it may become one if JSON-ID is used for large datasets. In any case, this may have to be documented somewhere because it may not be 100% obvious for a first-time reader (best practices document?).

---

That being said: one way of handling that over the Web is if the server returns

```
[
    {
        "@id" : "http://example.org/gr1",
        "@graph" : {
            "@id" : "a:something",
            "a:predicate" : "value1",
            "b:predicate" : "c:something1"
        }
    },
    {
        "@id" : "http://example.org/gr2",
        "@graph" : {
            "@id" : "a:something",
            "a:predicate" : "value2",
            "b:predicate" : "c:something2"
        }
    }
]
```

and the `@context` is returned via an HTTP Link. According to section 6, the processor is supposed to repeat the context for each element of the array...

pchampin:
... which unfortunately gives the wrong impression that there are three graphs in this dataset. And I'm assuming this is why Ivan didn't naturally came up with this solution.

Wouldn't the following also work?
```json
{
  "@context" : {
[...]

PR #208, opened 2019-07-30
Describe included blocks.

gkellogg:
For #19.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/208.html" title="Last updated on Aug 2, 2019, 5:28 PM UTC (809409a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/208/ab5050d...809409a.html" title="Last updated on Aug 2, 2019, 5:28 PM UTC (809409a)">Diff</a>

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -225,6 +225,9 @@
     A <a>named graph</a> created from the value of a <a>map entry</a>
     having an <a>expanded term definition</a>
     where <code>@container</code> is set to  <code>@graph</code>.</dd>
+  <dt class="changed"><dfn data-cite="JSON-LD11#dfn-included-block">included block</dfn></dt><dd class="changed">
+    An <a>included block</a> is an <a>entry</a> in a <a>node object</a> where the key is `@included` or an alias
+    and the value is one ore more <a>node objects</a>.</dd>
   <dt><dfn data-cite="JSON-LD11#dfn-index-map">index map</dfn></dt><dd>
     An <a>index map</a> is a <a>map</a> value of a <a>term</a>
     defined with <code>@container</code> set to <code>@index</code>,
--- a/common/terms.html
+++ b/common/terms.html
@@ -226,7 +226,7 @@
     having an <a>expanded term definition</a>
     where <code>@container</code> is set to  <code>@graph</code>.</dd>
   <dt class="changed"><dfn data-cite="JSON-LD11#dfn-included-block">included block</dfn></dt><dd class="changed">
-    An <a>included block</a> is an <a>entry</a> in a <a>node object</a> where the key is `@included` or an alias
+    An <a>included block</a> is an <a>entry</a> in a <a>node object</a> where the key is either `@included` or an alias of `@included`
     and the value is one ore more <a>node objects</a>.</dd>
   <dt><dfn data-cite="JSON-LD11#dfn-index-map">index map</dfn></dt><dd>
     An <a>index map</a> is a <a>map</a> value of a <a>term</a>
--- a/index.html
+++ b/index.html
@@ -608,6 +608,10 @@
         Used in a <a>context definition</a> to load an external context
         within which the containing <a>context definition</a> is merged.
         This can be useful to add JSON-LD 1.1 features to JSON-LD 1.0 contexts.</dd>
+      <dt class="changed">`@included`</dt><dd class="changed">
+        Used in in a top-level <a>node object</a> to define an <a>included block</a>,
+        for including secondary <a>node objects</a> within another <a>node object</a>.
+      </dd>
     </dl>
 
     <p>All keys, <a>keywords</a>, and values in JSON-LD are case-sensitive.</p>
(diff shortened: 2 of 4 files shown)


azaroth42:
LGTM! :)

azaroth42:
(Sorry, misclick)

common/terms.html
```diff
@@ -225,6 +225,9 @@
     A <a>named graph</a> created from the value of a <a>map entry</a>
     having an <a>expanded term definition</a>
     where <code>@container</code> is set to  <code>@graph</code>.</dd>
+  <dt class="changed"><dfn data-cite="JSON-LD11#dfn-included-block">included block</dfn></dt><dd class="changed">
+    An <a>included block</a> is an <a>entry</a> in a <a>node object</a> where the key is `@included` or an alias
```
dlongley:
```suggestion
    An <a>included block</a> is an <a>entry</a> in a <a>node object</a> where the key is either `@included` or an alias of `@included`
```

index.html
```diff
@@ -608,6 +608,10 @@ <h2>Syntax Tokens and Keywords</h2>
         Used in a <a>context definition</a> to load an external context
         within which the containing <a>context definition</a> is merged.
         This can be useful to add JSON-LD 1.1 features to JSON-LD 1.0 contexts.</dd>
+      <dt class="changed">`@included`</dt><dd class="changed">
+        Used in in a top-level <a>node object</a> to define an <a>included block</a>,
```
```suggestion
        Used in a top-level <a>node object</a> to define an <a>included block</a>,
```

index.html
```diff
@@ -8272,6 +8276,150 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
 </section>
 </section>
 
+<section class="informative changed"><h3>Included Nodes</h3>
+  <p>Sometimes it is useful to list node objects as part of another node object.
+    For instance, to represent a set of resources which are used by some other
+    resource.
+    In JSON-LD. these resources could all be contained as members of an array,
```
```suggestion
    In JSON-LD, these resources could all be contained as members of an array,
```

index.html
```diff
@@ -11918,6 +12073,16 @@ <h2>Type Maps</h2>
       added as a <code>@type</code> of the <a>node object</a> value when expanding.</p>
   </section>
 
+  <section class="changed">
+    <h2>Included Blocks</h2>
+
+    <p>An <a>included block</a> is used to provide a set of <a>node objects</a>.
+      An <a>included block</a> MAY appear as the value of a member of a node object with the key of `@included`, or an alias.
```
```suggestion
      An <a>included block</a> MAY appear as the value of a member of a node object with either the key of `@included` or an alias of `@included`.
```

index.html
```diff
@@ -12246,6 +12411,12 @@ <h2>Keywords</h2>
       This keyword is described further in <a class="sectionRef" href="#context-definitions"></a>,
       and <a class="sectionRef" href="#default-vocabulary"></a>.
     </dd>
+    <dt><code>@included</code></dt><dd>
+      The `@included` keyword MAY be aliased,
```
```suggestion
      The `@included` keyword MAY be aliased and
```

iherman:
Can we simplify the example to contain only what is really necessary to understand the feature? From the point of view of this feature aliasing `@type`, `@id`, or indeed `@included`, but also the usage of `@base` seems to be totally irrelevant. The same for the usage of type for the included objects. 

This would help the user concentrating on what is really essential in this case...

(Otherwise it looks o.k. to me...)

gkellogg:
The example came from Rob’s original use case, but I can simplify it easily enough.

iherman:
Actually... the comment of @dlongley in https://github.com/w3c/json-ld-syntax/issues/128#issuecomment-516936845 made me realize that, at least for me, the most important aspect of this feature is not what is in the proposed text. Indeed, what `@included` can replace is the misleading top level `@graph` when what one wants to expess is a “bush”, ie, several top level objects with their (recursive) properties. This is used all over the place in Turtle data, and for which the only solution in JSON-LD 1.0 is to use a top-level `@graph` property whoise value is... _not_ a graph in the dataset sense (certainly not the same way as when a TriG data is encoded in JSON-LD).

If my understanding is indeed correct, I would think that such an example should be “the” core example for this feature, and the cross references in @azaroth42’s example is just a welcome consequence... 

Or, at least, this fact should be emphasized in the text imho.

gkellogg:
@iherman Please suggest some text and an example to help flesh this out in the spec. I'll be happy to integrate it into the text.

dlongley:
@iherman,

> Indeed, what @included can replace is the misleading top level @graph when what one wants to expess is a “bush”, ie, several top level objects with their (recursive) properties.

Yes, this is also my understanding -- I left a comment to similar affect over on the API PR:

https://github.com/w3c/json-ld-api/pull/128#pullrequestreview-269051034

iherman:
@gkellogg 

> @iherman Please suggest some text and an example to help flesh this out in the spec. I'll be happy to integrate it into the text.

Hopefully, the PR over this PR works: https://github.com/w3c/json-ld-syntax/pull/209

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-08-02-json-ld#section4-1).

- `RESOLVED:  focus <code>@included</code> text and example on original inclusion use case; mention value of it as an <code>@graph</code> replacement for bushes–and reference primer for further reading`
- `RESOLVED:  close issue #19 with merger of <code>@included</code> related PRs`
4.1. Indexing with <code>@included</code><br/>
Ivan Herman: -> <a href="https://github.com/w3c/json-ld-syntax/pull/208">https://github.com/w3c/json-ld-syntax/pull/208</a> issue PR<br/>
Ivan Herman: -> <a href="https://github.com/w3c/json-ld-syntax/issues/19">https://github.com/w3c/json-ld-syntax/issues/19</a> issue itself<br/>
Benjamin Young: We made good progress on this last week.<br/>
Gregg Kellogg: It is <code>@included</code> now.16:13:56 &lt;bigbluehat> <a href="https://jsonapi.org/">https://jsonapi.org/</a><br/>
… <code>@included</code> comes from the JSON.API spec, and we are adopting this.<br/>
… Right now it’s just a container for collecting node objects that don’t have a direct rel with the node in which they are contained.<br/>
… There’s been some exchange on the issue highlighting a bush-like use for included.<br/>
… In JSON-LD 1.0 the top-level graph is used to collect nodes is a corner case. Everywhere else where graphs are used are seen as named graph.<br/>
… Included doesn’t carry that baggage.<br/>
… So <code>@included</code> can be used in favor of <code>@graph</code> in these places.<br/>
… In 1.0, you can’t have a graph name being a property of another node. With <code>@included</code> you can.<br/>
… We can’t use <code>@graph</code> to define a default graph.<br/>
… Except when it is the only property in a top-level object.<br/>
Benjamin Young: How does the actual inclusion take place?<br/>
Gregg Kellogg: The shape is similar to JSON.API. The value is seen as an array of node objects. If you have a node that is a value of a prop….<br/>
… Needed in jsonapi for node references that are not included in the main document as references, but should be included aside.<br/>
… Included blocks can be nested, and will be flattened out when done.<br/>
… You won’t compact back to included blocks after flattening.<br/>
… You can use included in a frame and have it match on diff subjects.<br/>
… The name <code>@included</code> is out of sync with other keywords.<br/>
… Dave suggested <code>@include</code><br/>
… Just like jsonapi<br/>
Ivan Herman: It’s becoming bikeshedding<br/>
… I would go further than what you did. Ex 1.1.1 and 1.1.2 (bushes) should be removed.<br/>
… We should convince people to not use those forms anymore.<br/>
Gregg Kellogg: <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/208.html#example-111-using-graph-to-explicitly-express-the-default-graph">https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/208.html#example-111-using-graph-to-explicitly-express-the-default-graph</a><br/>
Gregg Kellogg: By removing these we won’t lose anything. We would have to remove everything after the note.<br/>
Gregg Kellogg: <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/208.html#example-103-simple-data-with-several-top-level-nodes">https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/208.html#example-103-simple-data-with-several-top-level-nodes</a><br/>
Gregg Kellogg: Also, we may want to reverse example 103 and 104. To clarify writing bushes.<br/>
Ivan Herman: Referring to <code>@graph</code> is misleading, as it has not been explained yet there.<br/>
Gregg Kellogg: We may want to change an example in the <code>@graph</code> section then.<br/>
… We can also just leave that out and use it in the best practices document.<br/>
Ivan Herman: Yes<br/>
Benjamin Young: Flattened representation will still contain <code>@graph,</code> so readers will have to know what it does.<br/>
… This plumbing shift is significant.<br/>
… This is going to cause issues for the json people that are operating on the flattened form.<br/>
Gregg Kellogg: If you flatten with a context, it would introduce a graph to contain it. This would change the shape dramatically.<br/>
… Same with framing. In 1.1 we don’t use an <code>@graph</code> at the top level if not needed.<br/>
… We could change the algo to use included instead.<br/>
… But we may not want to do that.<br/>
… So do we want to replace the main usage of <code>@graph</code> to <code>@included?</code><br/>
… Included allows embedded nodes to go to one place. Like in jsonapi, they don’t want to include referenced things inline, but only a reference to an included block.<br/>
Benjamin Young: Useful for reducing payload size. And only including referenced things once.<br/>
… These are just IRI references?<br/>
Gregg Kellogg: There is no magic going on.<br/>
Ivan Herman: In JSON-LD it used to be hard to do these indexed references.<br/>
… The bush features can now be expressed in two different ways.<br/>
… These things happen.<br/>
… It’s a matter of taste which one you prefer.<br/>
… I personally always hated graph for representing bushes.<br/>
… I like this new included representation for bushes.<br/>
… I don’t want to hide the fact that bushes can be described with included instead of the graph ‘hack’.<br/>
Gregg Kellogg: We can say that included can also be used without other props in node object for describing node objects without semantic relationship.<br/>
Ivan Herman: I’m fine with that.<br/>
Benjamin Young: <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/208.html#included-blocks-to-be-flattened">https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/208.html#included-blocks-to-be-flattened</a><br/>
Benjamin Young: If <code>@included</code> were <code>@graph,</code> this would make a named graph?<br/>
Gregg Kellogg: I think this would make one or two named graphs.<br/>
Benjamin Young: With included it won’t make named graphs.<br/>
Gregg Kellogg: Yes, just objects.<br/>
Benjamin Young: I see the value, but not keen on the new keyword.<br/>
… I think we need to explain these next to each other, with their nuances.<br/>
… The initial reason for this feature was not meant to displace graph.<br/>
… It was meant to bring in other referenced objects in the document.<br/>
Gregg Kellogg: What jsonld always had was the ability to reference node<br/>
… by defining <code>@id</code> or <code>@vocab</code> you can define that thing.<br/>
… our mission is to use json in the wild where this is a pattern of usage<br/>
Benjamin Young: You said exactly what I was typing.<br/>
… it would be good to use an example from jsonapi<br/>
Gregg Kellogg: jsonapi examples are quite long, with a lot of nesting<br/>
… we have a test case from jsonapi<br/>
… may be too long for here. But may be good for best practices document.<br/>
… It would overly complicated the spec to include here.<br/>
Benjamin Young: This solves the jsonapi case by aliasing included to <code>@included.</code><br/>
Gregg Kellogg: Yes, you can have multiple properties that have multiple aliases.<br/>
… included can be a nested object<br/>
Ivan Herman: Can we talk about things that go to the primer?<br/>
Gregg Kellogg: What to do with example?<br/>
Ivan Herman: Switch the order of problem of Rob. In the primer we will have to spend more words on the fact that there are different things that can be used to do the same thing.<br/>
… We must have a primer.<br/>
… The current doc is already huge.<br/>
Benjamin Young: We need distinction in the main spec explaining diff between included and graph<br/>
Ivan Herman: To be honest, at this level there is no real diff between the examples.<br/>
… this is a side-effect with included.<br/>
… we should not fiddle around with that<br/>
Benjamin Young: The graph foundation exists in flattened output, and this won’t go away.<br/>
… this needs clarification<br/>
Gregg Kellogg: The use of included on its own is a by-product of the feature.<br/>
… it does not need its own description in the spec<br/>
… There are use cases where that can be useful<br/>
… That better lies in a non-normative text.<br/>
Benjamin Young: The bush usage would go to the primer<br/>
… focus of the text would go back to inter-document referencing.<br/>
Ivan Herman: We are falling back to other extreme that I don’t agree with<br/>
… we are hiding a feature of included<br/>
… In the inclusion part we should mention the alternative representation of bushes.<br/>
… because current graph-based bushes are a hack<br/>
… It has been around for a while, but we still should mention it.<br/>
Benjamin Young: This is an accidental feature since recently.<br/>
… it is essential to flattened output. I don’t see it as a hack.<br/>
… For non-turtle/trig users.<br/>
Ivan Herman: For semweb folks you don’t care it is a hack.<br/>
… we get into taste issues<br/>
… I don’t want to hide it.<br/>
Proposed resolution: focus <code>@included</code> text and example on original inclusion use case; mention value of it as an <code>@graph</code> replacement for bushes–and reference primer for further reading (Benjamin Young)<br/>
Ivan Herman: +0.5<br/>
Gregg Kellogg: +1<br/>
Benjamin Young: +1<br/>
Tim Cole: +1<br/>
Ruben Taelman: +1<br/>
Ivan Herman: If we have a primer, a reference can be put into it in CR<br/>
Resolution #2: focus <code>@included</code> text and example on original inclusion use case; mention value of it as an <code>@graph</code> replacement for bushes–and reference primer for further reading<br/>
Ivan Herman: include or included?<br/>
Benjamin Young: People may expect a URI to be included. But this is wrong.<br/>
Ruben Taelman: what was the keyword in JSON API?<br/>
Gregg Kellogg: included<br/>
Ruben Taelman: in that case, it might be helpful to be consistent with that<br/>
Tim Cole: +1 for <code>@included</code><br/>
Proposed resolution: close issue #19 with merger of <code>@included</code> related PRs (Benjamin Young)<br/>
Tim Cole: +1<br/>
Ivan Herman: +1<br/>
Ruben Taelman: +1<br/>
Benjamin Young: +1<br/>
Gregg Kellogg: +1<br/>
Resolution #3: close issue #19 with merger of <code>@included</code> related PRs<br/>
4<br/>
</details>

PR #209, opened 2019-08-01
Added the example for bushes

iherman:
(I hope this PR for a PR works:-)

I have made a change on the section of `@included` to emphasize the simple bush feature, but also maintaining the original example. I am sure the text needs harmonization as for the style.

I was also not sure about the paragraph on the JSON API; I have never used that thing, so I cannot comment on it. I am actually not sure it is worth keeping it here; it is not necessarily relevant imho. (or should be put as a note if we really want to keep to it).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/209.html" title="Last updated on Aug 1, 2019, 8:34 PM UTC (eea1017)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/209/4068a74...eea1017.html" title="Last updated on Aug 1, 2019, 8:34 PM UTC (eea1017)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -8279,7 +8279,7 @@ the data type to be specified explicitly with each piece of data.</p>
 <section class="informative changed"><h3>Included Nodes</h3>
   <p>
     Most of the examples so far were about a top level node with its properties referring to, possibly, other nodes in a tree-like fashion.
-    However, there are cases when the data should combine <em>several</em> such objects whithin one JSON-LD file.
+    However, there are cases when the data should combine <em>several</em> such objects within one JSON-LD document.
     In JSON-LD, these resources could all be contained as members of an array:
   </p>
 
--- a/index.html
+++ b/index.html
@@ -8277,22 +8277,134 @@ the data type to be specified explicitly with each piece of data.</p>
 </section>
 
 <section class="informative changed"><h3>Included Nodes</h3>
-  <p>Sometimes it is useful to list node objects as part of another node object.
+  <p>
+    Most of the examples so far were about a top level node with its properties referring to, possibly, other nodes in a tree-like fashion.
+    However, there are cases when the data should combine <em>several</em> such objects whithin one JSON-LD file.
+    In JSON-LD, these resources could all be contained as members of an array:
+  </p>
+
+  <pre class="example input" id="bush-in-array" title="Simple data with several top level nodes" data-transform="updateExample">
+  <!--
+[
+    {
+        "@id": "http://manu.sporny.org/about#manu",
+        "@type": "http://xmlns.com/foaf/0.1/Person",
+        "name": "Manu Sporny",
+        "knows": { "@id": "https://greggkellogg.net/foaf#me" }
+    }, {
+        "@id": "https://greggkellogg.net/foaf#me",
+        "@type": "http://xmlns.com/foaf/0.1/Person",
+        "name": "Gregg Kellogg",
+        "knows": { "@id": "http://manu.sporny.org/about#manu"}
+    }  
+]
+  -->    
+  </pre>
+
+  <p>
+    The problem with this approach is that we loose the ability to use common contexts.
+    To overcome this problem, <a>included blocks</a> may be used to collect such secondary <a>node objects</a>. 
+    Semantically, this is the same as if the node objects were embedded, or were contained in some enclosing <a>array</a>:
+  </p>
+
+  <aside class="example ds-selector-tabs changed"
+         title="Simple data with several top level nodes with a common context">
+    <div class="selectors">
+      <button class="selected" data-selects="compacted">Compacted (Input)</button>
+      <button data-selects="expanded">Expanded (Result)</button>
+      <button data-selects="statements">Statements</button>
+      <button data-selects="turtle">Turtle</button>
+      <a class="playground" target="_blank"></a>
+    </div>
+    <pre class="compacted input selected nohighlight" data-transform="updateExample">
+    <!--
+    {
+      "@context": {
+        "generatedAt": {
+          "@id": "http://www.w3.org/ns/prov#generatedAtTime",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "Person": "http://xmlns.com/foaf/0.1/Person",
+        "name": "http://xmlns.com/foaf/0.1/name",
+        "knows": {"@id": "http://xmlns.com/foaf/0.1/knows", "@type": "@id"}
+      },
+      "@included":**** [
+        {
+          "@id": "http://manu.sporny.org/about#manu",
+          "@type": "Person",
+          "name": "Manu Sporny",
+          "knows": "https://greggkellogg.net/foaf#me"
+        }, {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -8277,22 +8277,134 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
 
 <section class="informative changed"><h3>Included Nodes</h3>
-  <p>Sometimes it is useful to list node objects as part of another node object.
-    For instance, to represent a set of resources which are used by some other
-    resource.
-    In JSON-LD, these resources could all be contained as members of an array,
-    but this does not conveniently allow them to share a common context.</p>
+  <p>
+    Most of the examples so far were about a top level node with its properties referring to, possibly, other nodes in a tree-like fashion.
+    However, there are cases when the data should combine <em>several</em> such objects whithin one JSON-LD file.
```
```suggestion
    However, there are cases when the data should combine <em>several</em> such objects within one JSON-LD document.
```

index.html
```diff
@@ -8277,22 +8277,134 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
+]
+  -->    
+  </pre>
 
-  <p>In JSON-LD, <a>included blocks</a> may be used to collect such secondary <a>node objects</a>
-    which can be referenced from a primary <a>node object</a>.
-    Semantically, this is the same as if the node objects were embedded, or were contained
-    in some enclosing <a>array</a>.</p>
+  <p>
+    The problem with this approach is that we loose the ability to use common contexts.
```
```suggestion
    The problem with this approach is that we lose the ability to use a common context.
```

index.html
```diff
@@ -8277,22 +8277,134 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
+    To overcome this problem, <a>included blocks</a> may be used to collect such secondary <a>node objects</a>. 
+    Semantically, this is the same as if the node objects were embedded, or were contained in some enclosing <a>array</a>:
```
```suggestion
    Semantically, this is the same as if the node objects were embedded or were contained in some enclosing <a>array</a>:
```

iherman:
Thanks @dlongley !

index.html
```diff
@@ -8277,22 +8277,134 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
+]
+  -->    
+  </pre>
 
-  <p>In JSON-LD, <a>included blocks</a> may be used to collect such secondary <a>node objects</a>
-    which can be referenced from a primary <a>node object</a>.
-    Semantically, this is the same as if the node objects were embedded, or were contained
-    in some enclosing <a>array</a>.</p>
+  <p>
+    The problem with this approach is that we lose the ability to use a common context.
```
azaroth42: Is this true? This seems like the use case for `{"@context": {}, "@graph": [ ... ]}` ?

azaroth42: e.g.:  http://tinyurl.com/y54yasoh

iherman: Well, yes, the solution so far to the problem is to use `@graph`. But it turns out that `@included` can play exactly the same role, except that using that keyword (instead of `@graph`) is cleaner. The problem with using `@graph` for something like that is that it looks like we are talking about a dataset when, in fact, we are not!

B.t.w., and that is relevant, the example is almost exactly the same is a few chapter below (example 111 in the PR) where datasets/named graphs are described. But there is difference between named graphs (or data sets) and a single graph which can then be handled via `@included`.

B.t.w., and I forgot to do that: if we agree with this PR, then the text around 111 should be redone. Of course, 111 remains valid due to backward compatibility reasons but I would propose that the `@included` should be promoted instead.

gkellogg:
Looks good, I fixed the examples and added a note to Example 111.

BigBlueHat:
I don't think this was the intended use of `@included`...and I find using it for bushes confusing--it's yet another way to do them rather than a meaningful replacement for the others.

Bushes can be written (as I understand them):
```json
[{"@context": "...local context...", "@id": "...subject..."}, {}]
```
or
```json
{
  "@context": "...shared context...",
  "@graph": [{"@id": "...subject..."}, {}]
}
```

And datasets can be written as...
```json
{
  "@context": "...shared context...",
  "@id": "...name of graph...",
  "@graph": [{"@id": "...subject..."}, {}]
}
```

I've never found that confusing...but maybe I'm missing something?

BigBlueHat:
Put differently, if the following two examples are identical, we don't need the new one.

New `@included`-based "bush."
```json
{
  "@context": {
    "Person": "http://xmlns.com/foaf/0.1/Person",
    "name": "http://xmlns.com/foaf/0.1/name",
    "knows": {"@id": "http://xmlns.com/foaf/0.1/knows", "@type": "@id"}
  },
  "@included": [{
    "@id": "http://manu.sporny.org/about#manu",
    "@type": "Person",
    "name": "Manu Sporny",
    "knows": "https://greggkellogg.net/foaf#me"
  }, {
    "@id": "https://greggkellogg.net/foaf#me",
    "@type": "Person",
    "name": "Gregg Kellogg",
    "knows": "http://manu.sporny.org/about#manu"
  }]
}
```
vs. existing `@graph`-based bush...
```json
{
  "@context": {
    "Person": "http://xmlns.com/foaf/0.1/Person",
    "name": "http://xmlns.com/foaf/0.1/name",
    "knows": {"@id": "http://xmlns.com/foaf/0.1/knows", "@type": "@id"}
  },
  "@graph": [{
    "@id": "http://manu.sporny.org/about#manu",
    "@type": "Person",
    "name": "Manu Sporny",
    "knows": "https://greggkellogg.net/foaf#me"
  }, {
    "@id": "https://greggkellogg.net/foaf#me",
    "@type": "Person",
    "name": "Gregg Kellogg",
    "knows": "http://manu.sporny.org/about#manu"
  }]
}
```

@gkellogg do these generate the same graph/bush output?

iherman:
I think they are the same indeed, and if we have `@included` then indeed we have two ways of expressing the same.

As for using `@graph`: I do find it confusing. These two things are not the same, semantically (in Turtle and TriG, respectively):

```
a:b c:d "something" .
```

```
[] { a:b c:d "something". }
```

Using `@graph` for a bush, for me, would naturally map to the latter but, the way we use it, it maps in the former. That _is_ confusing in my view.

iherman:
I know the second example above is not really valid, though... A better one is to say

```
 a:b c:d "something".
<http://a.b.c.> {}
```

The point is: let us not mix up datasets and simple graphs...

gkellogg:
One difference is that `@graph` takes on a different meaning when other properties are present, which `@included` doesn't. This means you can have an `@included` block where the members are graph objects, each denoting a different named graph.

BigBlueHat:
```turtle
a:b c:d "something" .
```
becomes
```json
{
  "@graph": [{"@id": "a:b", "c:d": "something"}]
}
```

And the dataset one...
```turtle
<http://a.b.c.> { a:b c:d "something". }
```
becomes...
```json
{
  "@id": "http://a.b.c.",
  "@graph": [{"@id": "a:b", "c:d": "something"}]
}
```

I really don't see the confusion. A dataset is an `@graph` with an identifier.

Issue #225, opened 2019-09-16
Data Model Overview is misleading?


azaroth42:
> The graph contains nodes, which are connected by directed-arcs. A node is typically data such as a string, number, typed values (like dates and times) or an IRI.

Propose to use the more common "edges" rather than "arcs" and to remove the hyphen.  Also, that the IRI is a different node than the resource itself ... an IRI is the value of `@id`, whereas a JSON object representing a resource is (IMO) the most common node in the graph.

Propose:

> The graph contains nodes, which are connected by directed edges. A node is either a resource with properties, or the data values of those properties including strings, numbers, typed values (like dates and times) and IRIs.

gkellogg:
RDF Concepts calls them args: "An RDF graph can be visualized as a node and directed-arc diagram, in which each triple is represented as a node-arc-node link".

This text is essentially right out of RDF Concepts. Edges may be more common, but we use the RDF data model, and this is where it's defined.

Also, RDF Concepts does not really distinguish between an IRI and the node that it labels. But, I think we can use most of what you suggest by keeping "directed-arc".

> The graph contains nodes, which are connected by _directed-arcs_. A node is either a resource with properties, or the data values of those properties including strings, numbers, typed values (like dates and times) and IRIs.

azaroth42:
:+1: Works for me

PR #265, opened 2019-09-24
Azaroth suggestions

### LINKED ISSUE #217: Should abstract refer to 1.1 version?
Propose:
> This specification defines JSON-LD **1.1**, a JSON-based format to serialize Linked Data.

### LINKED ISSUE #218: Remove somewhat passive-aggressive "if desired" from Introduction
Propose:

> It is also designed to be usable as RDF ~, if desired, for use~ **in conjunction** with other Linked Data technologies like SPARQL ...

### LINKED ISSUE #219: default language is described rather than defined in 1.4
Currently in 1.4:
>The default language is set in the context using the @language key whose value MUST be a string representing a [BCP47] language code or null.

But that doesn't define what the default language is, just how to set it.

Propose (or similar):

> The default language is the language used when a string does not have a language associated with it directly. It can be set in the context [...]

### LINKED ISSUE #221: Typo: it's --> its in id map definition
> If a value in the id map contains a key expanding to @id, ~it's~ **its** value MUST be equivalent to the referencing key ...

### LINKED ISSUE #222: Typo: ore --> or in included block definition
> and the value is one ~ore~ **or** more node objects.

Also, propose:

> and the value is an array of one or more node objects.

To avoid the interpretation that a single node object is okay (which it isn't ... right?)

### LINKED ISSUE #223: Typo: it's --> its in type map definition
> If the value contains a term expanding to @type, ~it's~ **its** values are merged with the map value when expanding.

### LINKED ISSUE #224: Simplicity design goal can be even simpler!
Current wording in 1.5 is:
> Developers only need to know JSON and two keywords (@context and @id) to use the basic functionality in JSON-LD.

I would propose that it's even simpler than this, as `@id` can be aliased away.

> Developers not concerned with Linked Data only need to understand JSON, and know to include but ignore the `@context` property, to use the basic functionality of JSON-LD.

### LINKED ISSUE #226: Overview implies literals are blank nodes
Currently:
> Within a directed graph, nodes may be unnamed, i.e., not identified by an IRI or representing data such as strings or numbers. Such nodes are called blank nodes, [...] 

Which implies that:
* some nodes are unnamed
* unnamed nodes are not identified by IRIs or,
* unnamed nodes represent data
* all unnamed nodes are blank nodes

Therefore literals are blank nodes, which is not true.

Suggest removing the mention of literals, and clarifying that node here means resource.

### LINKED ISSUE #227: @nested keyword definition does not describe what it does
> Collects a set of nested properties within a node object.

The definition, like the others in the section, should start with "Used to ..."

Given that it is used in the context, I propose something like:

> "Used to define a property of a node object that groups together properties of that node, but is not an edge in the graph."

### LINKED ISSUE #228: Definition of @none could be more specific
@none is used when the indexed node does not have the particular feature being indexed.

Propose, or similar:

> Used as an index value [...] used to index into other values, and the indexed node does not have the feature being indexed.

### LINKED ISSUE #229: Typo: sccope for scope in definition of @propagate
> ... change the ~sccope~ **scope** of that context.

### LINKED ISSUE #230: Potential for confusion of @version syntax?
The definition of `@version` says: 
> [...] available when processing mode has been explicitly set to `json-ld-1.1`.

*I* know what this means, but reading it as a newcomer I might be tempted to put the string `"json-ld-1.1"` as the value of `@version` rather than just `1.1`

I think this should be made clearer in the definition of the keyword?

### LINKED ISSUE #232: Refer to privacy / security concern of context deref in 3.1?
> In JSON-LD documents, contexts may also be specified inline. This has the advantage that documents can be processed even in the absence of a connection to the Web. Ultimately, this is a modeling decision and different use cases may require different handling.

I think this should call out the privacy and security concerns raised by @danbri et al. It is a modeling decision, but it's not *just* a modeling decision.

### LINKED ISSUE #233: Links in TR/ refer to github.io version not internal anchors
A bug with respec, or echidna?

### LINKED ISSUE #234: "List Objects and Set objects" ... ?
In section 3.4 there is a section heading "List Objects and Set objects" without any content.

Content should be supplied or the heading removed. Also "Set Objects" for consistency.

### LINKED ISSUE #235: Nested properties in section 4
Currently:
> to use an intermediate object to represent the properties

Propose to use "group" rather than represent ...

> Another JSON idiom often found in APIs is to use an intermediate object to group together related properties of an object; [...]

### LINKED ISSUE #236: Forward ref in 4.1 to 4.1.7-10 ?
In the intro of 4.1 there is quite some time spent describing the last-in wins approach to contexts.  In 1.1 this is no longer the whole picture, with the addition of `@import`, `@propagate`, scoped contexts and `@protected`

Propose a forward reference at the end of the text after example 20 to these sections.

### LINKED ISSUE #237: Text before example 25 should describe the result
That `Restaurant` is expanded to `http://example.com/vocab/Restaurant`.  Otherwise you have to remember to click through on to the expanded tab to see what the text is talking about.

### LINKED ISSUE #238: 4.1.2.1 should come after 4.1.3
As 4.1.3 introduces `@base`, but 4.1.2.1 uses it.

### LINKED ISSUE #239: Typo: it's --> its in 4.1.4
> both because it is defined with an expanded term definition, and because ~it's~ **its** `@id` does not end in a gen-delim character.

### LINKED ISSUE #240: Bad grammar in intro to 4.1.7
> This allows values to use term definitions, base IRI, vocabulary mapping or default language which is different from the node object they are contained in, as if the context was specified within the value itself.

The list mixes singular and plural, making the "which is different" read very badly. Propose to pluralize all of the options and use "are" instead.

### LINKED ISSUE #241: Move text inside example 46 to immediately before it
> This example, shows how properties and types can define their own scoped contexts, which are included when expanding.

This text should be before the example, not within the example block.

And then split the example into two, with the inner text between them.

### LINKED ISSUE #242: redefine / redefinition doesn't need hyphen
In 4.1.7 (and perhaps elsewhere) "re-defined" has a hyphen, but is a valid English word without it:  https://www.merriam-webster.com/dictionary/redefine

Similarly re-definition later in the note, and perhaps other composites.

### LINKED ISSUE #243: Grammar in 4.1.8
> with the exception of type-scoped contexts, which limits the affect of that context until the next node object is entered.

Propose:

> with the exception of type-scoped contexts which limit the effect of that context to until the next node object is entered.

### LINKED ISSUE #244: Typo and confusion in 4.1.9 intro
> The following examples illustrate how @import can be used to express a ~for~ type-scoped context that loads an imported context and sets @propagate to true and how to make similar modifications.

Propose:

> The following examples illustrate how @import can be used to express a type-scoped context that loads an imported context and sets @propagate to true **, as a technique for** how to make **other,** similar modifications.

### LINKED ISSUE #245: Typo: effect -> affect after example 50
>  or set other context-wide keywords that may effect how the imported context term definitions

Should be the verb homonym:

>  or set other context-wide keywords that may **affect** how the imported context term definitions

("effect" following example 51 is correct)

(Aren't you glad that my mother was an English teacher and professional proof-reader in the publishing industry ;) )

### LINKED ISSUE #247: Note in 4.2.1 should clarify not setting node type via context
While the note refers to 3.5, neither section are explicit that you cannot set the class of a node object by assigning a value to `@type` in the context.  We have had so many misunderstandings about this, it would be good to have it as explicit as possible to refer people to when they come to ask about it.

### LINKED ISSUE #248: Example 63 is still hard to understand
I know that this was the original example, as submitted, but the use of "fred" as a property rather than a resource makes it more complicated than necessary to see what is going on.

Changing it slightly to follow the previous examples of social networking would help with comprehension:

```
{
  "@context": {
    "@base": "http://example1.com/",
    "@vocab": "http://example2.com/",
    "knows": {"@type": "@vocab"}
  },
  "knows": [
    {"@id": "barney", "name": "barney"},
    "barney"
  ]
}
```

And similar through to example 65

### LINKED ISSUE #249: Note at end of 4.2.3 is out of date
It reads as follows, which should be deleted as it's now not allowed:

> [...] It also allows a compact IRI (or even an absolute IRI) to be defined as something else entirely. For example, one could specify that the term http://example.org/zoo should expand to http://example.org/river, but this usage is discouraged because it would lead to a great deal of confusion among developers attempting to understand the JSON-LD document.

### LINKED ISSUE #250: example 67 uses not best practice of node embedded context
And instead I think we would now recommend to use a scoped context, no?

```
{
  "@context": {
    ...
    "@language": "ja"
  },
  "name": "花澄",
  "details": {
    "@context": {
      "@language": null
    },
    "occupation": "Ninja"
  }
}
```

Would become

```
{
  "@context": {
    ...
    "details": {
      "@context": {
        "@language": null
      }
    },
    "@language": "ja"
  },
  "name": "花澄",
  "details": {
    "occupation": "Ninja"
  }
}
```

This pattern should at least be demonstrated, even if 67 isn't changed

### LINKED ISSUE #251: 4.2.4 or 4.6.2 to note challenge of "en-us" and dot-property notation code
Both descriptions of the language map pattern talk about the utility of using object.en as a method of accessing the data. However, neither mention the subsequent headache of "-region" or other BCP47 values beyond the simple two letter code for the language as a whole.

If I use "en-us" as the language of my string then I just broke everyone's dot-property notation code, as they'll try in vain to do:

```
text = object.occupation.en - us;
```
and hopefully get a 'Variable Not Defined: us' error, and at worst (if us is defined as something that can be subtracted from a string) some very badly broken content.

This should be in a note in one or both of the descriptions.

### LINKED ISSUE #252: Delete type coercion reference from 4.5
Propose deleting:

> Note that type coercion of the knows property is not required, as the value is not a string.

As meaningless. It can't be type coerced at all, let alone not being required!

gkellogg:
* "This specification defines JSON-LD" => "This specification defines JSON-LD 1.1". Fixes #217. 
* Remove "if desired" Fixes #218.
* Improve _default language_ definition. Fixes #219.
* it's => its fixes. Fixes #221, fixes #223, and fixes #239.
* ore => or. Fixes #222. 
* Simplify simplicity. Fixes #224.
* Improve data model overview. For #225.
* Clarify that literals are not blank nodes. Fixes #226.
* Improve definition of `@nest`. Fixes #227.
* Update `@none` definition. Fixes #228.
* sccope => scope. Fixes #229.
* Clarify that `@version` is set to `1.1`, not `"json-ld-1.1"`. Fixes #230.
* Improve requirements of terms. Improve requirements of terms.
* Reference security considerations for remote contexts. Fixes #232.
* Update local dfns to not be links by removing data-cite. Fixes #233.
* Add a missing description of list objects and set objects. Fixes #234.
* Fix wording of nested properties. Fixes #235.
* Add forward references for other things that affect the context in section 4.1. Fixes #236.
* Better describe example 25. Fixes #237.
* Move subsection on using the document base for the default vocabulary until after the section on the document base (4.1.2.1 => 4.1.4). Fixes #238.
* Fix grammar in Scoped Contexts section. Fixes #240.
* Split example 46 into separate examples. Fixes #241.
* Re-defined => Redefined. Fixes #242.
* limits => limit. Fixes #243.
* Fix grammar in Imported Contexts section. Fixes #244.
* effect => affect. Fixes #245.
* Further clarify that type coercion takes place only on strings, not maps. Fixes #247.
* Make examples clearer. Fixes #248.
* Remove old warning in note about redefining IRIs to have a different meaning, which is now disallowed. Fixes #249.
* Use a scoped, rather than an embedded context for default language example. Fixes #250.
* Note that indexing of language tags depends on their not using sub-tags other than the primary language-tag. Fixes #251.
* Remove uninformative note on type coercion from 4.5. Fixes #252.
*

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/265.html" title="Last updated on Sep 24, 2019, 9:00 PM UTC (09dcd19)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/265/fecd5c4...09dcd19.html" title="Last updated on Sep 24, 2019, 9:00 PM UTC (09dcd19)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6500,9 +6500,7 @@ the data type to be specified explicitly with each piece of data.</p>
   </aside>
 
   <p>A <a>node object</a>, like the one used above, may be used in
-    any value position in the body of a JSON-LD document. Note that <a
-    href="#type-coercion">type coercion</a> of the <code>knows</code> property
-    is not required, as the value is not a string.</p>
+    any value position in the body of a JSON-LD document.</p>
 
   <p>While it is considered a best practice to identify nodes in a graph,
     at times this is impractical. In the data model, nodes without an explicit
--- a/index.html
+++ b/index.html
@@ -2947,7 +2947,7 @@
 
   <p class="changed">In this case, the <em>property</em> term would not normally be usable as a prefix, both
     because it is defined with an <a>expanded term definition</a>, and because
-    it's <code>@id</code> does not end in a
+    its <code>@id</code> does not end in a
     <a data-cite="RFC3986#section-2.2">gen-delim</a> character. Adding
     <code>"@prefix": true</code> allows it to be used as the prefix portion of
     the <a>compact IRI</a> <em>property:One</em>.</p>
@@ -12056,7 +12056,7 @@ the data type to be specified explicitly with each piece of data.</p>
       or a <a>term</a> which expands to <code>@none</code>,
       and the values MUST be <a>node objects</a>.</p>
 
-    <p>If the value contains a property expanding to <code>@id</code>, it's value MUST
+    <p>If the value contains a property expanding to <code>@id</code>, its value MUST
       be equivalent to the referencing key. Otherwise, the property from the value is used as
       the <code>@id</code> of the <a>node object</a> value when expanding.</p>
 
@@ -12082,7 +12082,7 @@ the data type to be specified explicitly with each piece of data.</p>
       and the values MUST be <a>node objects</a>
       or <a>strings</a> which expand to <a>node objects</a>.</p>
 
-    <p>If the value contains a property expanding to <code>@type</code>, and it's value
+    <p>If the value contains a property expanding to <code>@type</code>, and its value
       is contains the referencing key after suitable expansion of both the referencing key
       and the value, then the <a>node object</a> already contains the type. Otherwise, the property from the value is
       added as a <code>@type</code> of the <a>node object</a> value when expanding.</p>
--- a/index.html
+++ b/index.html
@@ -4694,9 +4694,10 @@ the data type to be specified explicitly with each piece of data.</p>
     "@context": {
       "@base": "http://example1.com/",
       "@vocab": "http://example2.com/",
-      "fred": {"@type": "@vocab"}
+      "knows": {"@type": "@vocab"}
     },
-    "fred": [
+    "@id": "fred",
+    "knows": [
       {"@id": "barney", "mnemonic": "the sidekick"},
       "barney"
     ]
@@ -4708,15 +4709,13 @@ the data type to be specified explicitly with each piece of data.</p>
        data-result-for="Term expansion for values, not identifiers-compacted">
   <!--
   [{
-    "http://example2.com/fred": [
-      {
-        "@id": "http://example1.com/barney",
-        "http://example2.com/mnemonic": [{"@value": "the sidekick"}]
-      },
-      {
-        "@id": "http://example2.com/barney"
-      }
-    ]
+    "@id": "http://example1.com/fred",
+    "http://example2.com/knows": [{
+      "@id": "http://example1.com/barney",
+      "http://example2.com/mnemonic": [{"@value": "the sidekick"}]
+    }, {
+      "@id": "http://example2.com/barney"
+    }]
   }]
   -->
   </pre>
@@ -4726,8 +4725,8 @@ the data type to be specified explicitly with each piece of data.</p>
     <thead><tr><th>Subject</th><th>Property</th><th>Value</th></tr></thead>
     <tbody>
       <tr><td>http://example1.com/barney</td><td>http://example2.com/mnemonic</td><td>the sidekick</td></tr>
-      <tr><td>_:b0</td><td>http://example2.com/fred</td><td>http://example1.com/barney</td></tr>
-      <tr><td>_:b0</td><td>http://example2.com/fred</td><td>http://example2.com/barney</td></tr>
+      <tr><td>http://example1.com/fred</td><td>http://example2.com/knows</td><td>http://example1.com/barney</td></tr>
+      <tr><td>http://example1.com/fred</td><td>http://example2.com/knows</td><td>http://example2.com/barney</td></tr>
     </tbody>
   </table>
   <pre class="turtle"
@@ -4742,7 +4741,7 @@ the data type to be specified explicitly with each piece of data.</p>
 
   ex1:barney ex2:mnemonic "the sidekick" .
 
-  [ ex2:fred ex2:barney, ex1:barney] .
+  ex1:fred ex2:knows ex1:barney, ex2:barney .
   -->
   </pre>
 </aside>
@@ -4765,7 +4764,7 @@ the data type to be specified explicitly with each piece of data.</p>
   of <code>@vocab</code> illustrates the behavior of interpreting "barney" relative to the document:</p>
 
 
-<aside class="example ds-selector-tabs changed"
+<aside class="example ds-selector-tabs"
        title="Terms not expanded when document-relative">
   <div class="selectors">
     <button class="selected" data-selects="compacted">Compacted (Input)</button>
@@ -4780,9 +4779,10 @@ the data type to be specified explicitly with each piece of data.</p>
     "@context": {
       "@base": "http://example1.com/",
       "@vocab": "http://example2.com/",
-      "fred": {"@type": "@id"}
+      "knows": {"@type": "@id"}
     },
-    "fred": [
+    "@id": "fred",
+    "knows": [
       {"@id": "barney", "mnemonic": "the sidekick"},
       "barney"
     ]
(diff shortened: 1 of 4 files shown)


gkellogg:
Actually, there are several more I'll add on to this PR tomorrow.

gkellogg:
Still a couple of more to go.

gkellogg:
I'll do the rest in another PR.

index.html
```diff
@@ -805,10 +807,13 @@ <h2>The Context</h2>
...
-      <a>IRIs</a>. <a>Terms</a> are case sensitive
-      and any valid <a>string</a> that is not a reserved JSON-LD <a>keyword</a>
-      can be used as a <a>term</a>.</p>
+    <p>Simply speaking, a <a>context</a> is used to map <a>terms</a> to <a>IRIs</a>.
+      <a>Terms</a> are case sensitive and most any valid <a>string</a> that is not a reserved JSON-LD <a>keyword</a>
+      can be used as a <a>term</a>.
+      <span class="changed">The empty string `""` and strings that have the form
+        of a keyword (e.g., staring with `"@"`) must not be used as terms.
+        Strings that have the form of
+        an absolute IRI (e.g., containing a `":"`) should not be used as terms.</span></p>
```
pchampin: > Strings that [contain a colon] should not be used as terms.

Well, it is acceptable to have a *term definition* of the form `"foaf:knows": {"@type": "@id"}`.
Isn't `foaf:knows` a term in that case?

pchampin: My bad, the key for a term definition is not necessarily a term. Slightly confusing, but ok :)

gkellogg:
We can probably improve things more, but yes, the key for an expanded definition my be an IRI, which can’t be a term.

gkellogg:
Perhaps we should update the context definition to say that keys can be noted keywords and strings treated as terms of IRIs. The values of which may be strings, interpreted as IRIs, or expanded term definitions, or words to that effect.

Issue #246, opened 2019-09-16
keyword aliases for @id and @type cannot be singly @protected?


azaroth42:
In section 4.1.5 (and equivalent in the algorithm) it says that aliases applied to keywords are ignored, other than (per 4.3.3) the use of `@container:@set` applied to `@type`.

However, the `@protected` keyword can be applied to keyword aliases by setting it in the context overall. This is at odds with the text, and prevents protecting ONLY the (very common, and important) aliases for `@id` and `@type`. 

Propose (I know, 11th hour) that `@protected` be allowed as an entry for keyword aliases (unless we already discussed this and decided against it, and I'm just forgetting). We can already protect them, just by the syntactic sugar form of setting it at the context level.

azaroth42:
Works as expected:
```
{
  "@context": {
    "@protected": true,
    "id": "@id"
  },
  "id": "http://example.com/1"
}
```

Does not work, as `@protected` is ignored for aliases of keywords:

```
{
  "@context": {
    "id": {"@id": "@id", "@protected": true}
  },
  "id": "http://example.com/1"
}
```

azaroth42:
Also, the syntax is inconsistent with the API definition.

Syntax says that the definitions will be ignored, whereas API says "Any other value means that a keyword redefinition error has been detected and processing is aborted." 

The API is the correct intention, and the syntax document should be updated to this effect.

azaroth42:
Changing to editorial as the algorithm is in -api, not -syntax

azaroth42:
@gkellogg  to look through other normative sections to ensure consistency

azaroth42:
Resolution: Allow the use of `@protected` as a valid keyword in term definitions which are aliases of keywords or where the term is `@type`.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-09-19-json-ld#section3-2).

- `RESOLVED:  Allow the use of <code>@protected</code> as a valid keyword in term definitions which are aliases of keywords or where the term is <code>@type</code>`
<details><summary><i class="differentiate">View the transcript</i></summary>
keyword alias for <code>@id</code> and <code>@type</code> cannot be singly <code>@protected</code><br/>
Benjamin Young: See Syntax issue 246<br/>
Rob Sanderson: if you alias <code>@type</code> to type it can still be a set<br/>
[...]

PR #272, opened 2019-09-27
Improve description of keyword aliases using `@protected`

gkellogg:
and add more to the grammar.

For #246.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/272.html" title="Last updated on Sep 28, 2019, 5:35 PM UTC (eaa444e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/272/d28d3de...eaa444e.html" title="Last updated on Sep 28, 2019, 5:35 PM UTC (eaa444e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3055,12 +3055,19 @@
   <p><span class="changed">Other than for <code>@type</code>,</span> properties of
     <a>expanded term definitions</a> where the term is a <a>keyword</a>
     result in an error.
-    <span class="changed">When processing mode is set to <code>json-ld-1.1</code>,
-      a keyword may be used with an <a>expanded term definition</a> having
-      only the <a>entry</a> `@protected`;
-      see <a href="#protected-term-definitions" class="sectionRef"></a> for further details.
-      There is also an exception for <code>@type</code>;
+    <span class="changed">When processing mode is set to <code>json-ld-1.1</code>,t
+      there is also an exception for <code>@type</code>;
       see <a href="#using-set-with-type" class="sectionRef"></a> for further details.</span></p>
+
+  <p class="changed">When processing mode is set to <code>json-ld-1.1</code>,
+    aliases of <a>keywords</a> are either <a>simple term definitions</a>,
+    where the value is a <a>keyword</a>,
+    or a <a>expanded term definitions</a> with an `@id` <a>entry</a> and optionally an `@protected` <a>entry</a>;
+    no other entries are allowed.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -12050,9 +12057,14 @@ <h2>Context Definitions</h2>
     <code>@container</code>,
     <code class="changed">@context</code>,
     <code class="changed">@prefix</code>,
+    <code class="changed">@propagate</code>, or
```
pchampin: I did remove `@propagate` from this list earlier (b4dae4d), because I don't see anything the API where it is expected in an expanded term definition...

Am I missing something?

pchampin: I did remove `@propagate` from this list earlier (b4dae4d), because I don't see anything the API where it is expected in an expanded term definition...

Am I missing something?

index.html
```diff
@@ -12110,13 +12122,21 @@ <h2>Context Definitions</h2>
   <p class="changed">If an <a>expanded term definition</a> has an <code>@context</code> <a>entry</a>,
     it MUST be a valid <code>context definition</code>.</p>
 
+  <p class="changed">If a <a>context</a> contains the <code>@import</code>
```
pchampin: This whole paragraph is a copy of the one line 12017... Was the intent to replace "context" by "expanded term definition"?? That would be strange as `@import` can not appear in an expanded term definition...

index.html
```diff
@@ -12110,13 +12122,21 @@ <h2>Context Definitions</h2>
...
+    include an `@import` key, itself.</p>
+
   <p class="changed">If the <a>expanded term definition</a> contains the <code>@nest</code>
     <a>keyword</a>, its value MUST be either <code>@nest</code>, or a term
     which expands to <code>@nest</code>.</p>
 
   <p class="changed">If the <a>expanded term definition</a> contains the <code>@prefix</code>
     <a>keyword</a>, its value MUST be <code>true</code> or <code>false</code>.</p>
 
+  <p class="changed">If the <a>expanded term definition</a> contains the <code>@propagate</code>
```
pchampin: cf. my comment above, I don't think that an expanded term definition can contain the `@propagate` keyword...

index.html
gkellogg: No, I you’re right, only a context definition can contain `@propagate`.

gkellogg: No, I you’re right, only a context definition can contain `@propagate`.

Issue #11, opened 2018-06-30
Introducing @dir


gkellogg:
In some situations it is important/necessary to include the base direction of a text, alongside its language; see the [“Requirements for Language and Direction Metadata in Data Formats”](https://w3c.github.io/string-meta/) for further details. In practice, in a vanilla JSON, it would require something like:

  ```
  "title": [ { "value": "Moby Dick", "lang": "en" },
	         { "value": "موبي ديك", "lang": "ar"  "dir": "rtl"}
	       ]  
  ```

  (the example comes from that document).

  At this moment, I believe the only way you can reasonably express that in JSON-LD is via cheating a bit:

  ```
  "title": [ { "@value": "Moby Dick", "@language": "en" },
	         { "@value": "موبي ديك",  "@language": "ar"  "dir": "rtl"}
	       ]  
  ```

  and making sure that the `dir` term is _not_ defined in the relevant `@context` so that, when generating the RDF output, that term is simply ignored. But that also means that there is no round-tripping, that term will disappear after expansion.

  The difficulty lies in the RDF layer, in fact; RDF does not have any means (alas!) to express text direction. On the other hand, this missing feature is a general I18N problem whenever JSON-LD is used (there were issues when developing the Web Annotation Model, these issues are popping up in the Web Publication work, etc.).

  Here is what I would propose as a non-complete solution

  1. Let us introduce a `@dir` term, alongside `@language`. This means this term can be used in place of `dir` above, ie, it is a bona-fide part of a string representation, and would therefore be kept in the compaction/expansion steps, can also be used for framing.
  2. In JSON-LD 1.1, `@dir` is _ignored_ when transforming into RDF. I.e., only the language tag would be used.
  3. We _may_ initiate some work in the RDF community to solve this issue. There may be several ways, each of them require the RDF community to chime in
  	3.1. Define a mechanism of "parametrized" standard datatypes that represent a (language,direction) pair. One would then get something like`[] ex:title "موبي ديك"^^rdf:internationalText(ar,rtl) ;`
  	3.2. Go for a "generalized" RDF where strings can also appear as subjects (that has been a matter of dispute for a long time...). That would give the possibility to add such attribute to texts like directions
  	3.3. Some other mechanisms that I cannot think about
  4. In a future JSON-LD 1.* the `@dir` value can be properly mapped onto an RDF representing the right choices (if such choices are worked out)

[...]

PR #276, opened 2019-10-01
Describing `@direction`

gkellogg:
Note that examples won't build properly until it's fully supported in the Ruby implementation.

For #11.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/276.html" title="Last updated on Oct 10, 2019, 9:39 PM UTC (736d67f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/276/3a9dbec...736d67f.html" title="Last updated on Oct 10, 2019, 9:39 PM UTC (736d67f)">Diff</a>


gkellogg:
cc/ @r12a and @aphillips. The text references [String Meta](https://www.w3.org/TR/string-meta/) and riffs off of that content. It likely also may cause you to want to update that document based on the direction (*sic*) laid out here.

aphillips:
@gkellogg Thanks! I'll have a look at this PR shortly. String-Meta definitely is in need of an update per our various conversations; I'll try to start in on that as well.

index.html
```diff
@@ -12598,7 +12598,7 @@ <h2>Relationship to RDF</h2>
     <li class="changed">As an extension to the <a data-cite="RDF11-CONCEPTS#data-model">RDF data model</a>,
       <em>typed literals</em>, where the datatype is `xsd:string`, or <a>language-tagged strings</a>
       MAY include a <a>text direction</a>.
-      As there is not yet any standardized mechanism for representing the text direction
+      As there is currently no standardized mechanism for representing the <a>text direction</a>
       of <a>RDF literals</a>, the JSON-LD to RDF transformation can loose this direction.
```
aphillips: loose -> lose?

Or perhaps this should read more like ".. the JSON-LD to RDF transformation generally loses the base direction..."

index.html
```diff
@@ -5008,7 +5008,7 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
   </aside>
 
   <p>The example above would associate the <code>ja</code> language
-    code with the two <a>strings</a> <em>花澄</em> and <em>科学者</em>
+    code with the two <a>strings</a> <em lang="ja">花澄</em> and <em lang="ja">科学者</em>
     <a data-cite="BCP47#section-2">Languages codes</a> are defined in [[BCP47]].
```
aphillips: Nit: the term "language code" is generally associated with ISO 639. The term "language tag" (or subtag, in the case of elements of a tag) is usually used for BCP47 values.

index.html
```diff
@@ -5064,7 +5064,7 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
 
   <p>The example above would associate <em>忍者</em> with the specified default
     language code <code>ja</code>, <em>Ninja</em> with the language code
-    <code>en</code>, and <em>Nindža</em> with the language code <code>cs</code>.
+    <code>en</code>, and <em lang="cs">Nindža</em> with the language code <code>cs</code>.
     The value of <code>name</code>, <em>Yagyū Muneyoshi</em> wouldn't be
     associated with any language code since <code>@language</code> was reset to
     <a>null</a> in the <a>expanded term definition</a>.</p>
```
aphillips: you added only the one `lang` attribute (for Czech). Did you mean to add others here?

index.html
gkellogg: Fixed in 9b1e522554ac86b8079f21b10ca254a2c478ec2a.

index.html
gkellogg: Fixed in 9b1e522554ac86b8079f21b10ca254a2c478ec2a.

index.html
gkellogg: Fixed in 9b1e522554ac86b8079f21b10ca254a2c478ec2a.

gkellogg:
> * I think example 76 is incomplete: `@direction` should either be set to "ltr" or to null for the author; at present, it would set an "rtl" on an english text.

👍 

> * We should get the reaction of @aphillips or @r12a, but I believe the official terminology is "base direction". Maybe we should adopt it.

👍 

> * Where do you intend to put the two RDF mapping possibilities? The current document just says that the direction will be dropped, which is of course one option.

Those details go in the API document. The text references the [Deserialize JSON-LD to RDF Algorithm](https://www.w3.org/TR/json-ld11-api/#deserialize-json-ld-to-rdf-algorithm) for a description of how base direction can be preserved using non-standardized mechanisms. Do you think we need to detail them in the syntax document?

> * The text under example 75 got it wrong for the display of the Arabic title; it is not correctly displayed (I let @r12a give you the correct HTML encoding. My experience is that putting such texts into a separate paragraph that is set to rtl is safe; it would actually put the title into a separate, right-justified line, which may make things clearer)

Other than the use of `<em>` vs `<span>` it's exactly the same markup as is used in the string-meta document. If I change it to a `<div>` it works okay, but messes up the flow. Perhaps @aphillips or @r12a can suggest a remedy the preserves flow.

Interestingly, if I add exactly the same markup here, it seems to also mess up (but differently). <em lang="ar" dir="rtl">HTML و CSS: تصميم و إنشاء مواقع الويب</em>. Adding it to a new paragraph preserves order okay:

<em lang="ar" dir="rtl">HTML و CSS: تصميم و إنشاء مواقع الويب</em>

aphillips:
> Interestingly, if I add exactly the same markup here, it seems to also mess up (but differently). HTML و CSS: تصميم و إنشاء مواقع الويب. Adding it to a new paragraph preserves order okay:

> HTML و CSS: تصميم و إنشاء مواقع الويب

Regarding "it seems to mess up": the wrapping behavior is actually correct, if you think in terms of being an Arabic speaker--in an RTL language, the *start* of the string is on the *right side* (and notice the colon goes to the left of "CSS"). When I saw that this happened in the preview, I thought it was amusing (since it illustrates the need for bidi support) but also confusing if you don't understand what's going on. This is one reason why we tend to put the examples on lines by themselves (for clarity to folks with "left-to-right eyes"): we have to explain that the "brokenness" is because left-to-right readers are trained to expect the LTR token to start the line ("because it is next")--it's a form of confusion about where "start" is.

Ultimately, what you're experiencing is kind of the point of all this. Text that was previously broken in Arabic now reads correctly. It's instructive to have the reverse experience, I think. Here's a google translate of the "Interestingly" quote at the start of this comment (with an LTR span around the title):

> <p dir=rtl>ومن المثير للاهتمام ، إذا أضفت نفس الترميز تمامًا هنا ، فيبدو أن الأمر أيضًا يفسد (ولكن بشكل مختلف). <span dir=ltr>HTML و CSS: web design and creation</span>. إضافة إلى فقرة جديدة يحافظ على النظام بخير</p>

image


Anyway, use one line for legibility...

gkellogg:
@aphillips I found in your document in [4.1.1](https://www.w3.org/TR/string-meta/#how-it-works), the string is represented using `<span lang="ar" dir="rtl">HTML و CSS: تصميم و إنشاء مواقع الويب</span>` (<em lang="ar" dir="rtl">HTML و CSS: تصميم و إنشاء مواقع الويب</em>), which shows up to me with the Arabic part, followed by ":CSS , HTML". But, in https://github.com/w3c/json-ld-syntax/pull/276#issuecomment-537634650, the exact same span shows as  "HTML" followed by the Arabic part, followed by ":CSS ," So, the exact same span renders differently in different contexts.

But, if you think what we have in our doc looks correct, then I won't worry about it.

iherman:
> But, if you think what we have in our doc looks correct, then I won't worry about it.

I ran in the same problem in the [PR](https://github.com/w3c/pub-manifest/pull/92) on the publication manifest. I think that, for readability's sake, putting the problematic sentences into a separate paragraph is better. This is what I did in, e.g., [section 2.4.4.1](https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/92.html#value-localizable-string] and [section 2.6.2](https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/92.html#manifest-lang-dir-local). I agree with @aphillips that keeping such examples on separate lines is more readable

iherman:
> Those details go in the API document. The text references the Deserialize JSON-LD to RDF Algorithm for a description of how base direction can be preserved using non-standardized mechanisms. Do you think we need to detail them in the syntax document?

I am not sure. On the one hand, it would probably make the syntax doc too convoluted, on the other hand, as I said many times, the JSON-LD _user_ will tend to read the syntax document only. I guess once the API document is done in this respect, at the minimum a reference would be a good idea, but I could also see the value of having an informative appendix in the syntax document.

gkellogg:
> > But, if you think what we have in our doc looks correct, then I won't worry about it.
> 
> I ran in the same problem in the [PR](https://github.com/w3c/pub-manifest/pull/92) on the publication manifest. I think that, for readability's sake, putting the problematic sentences into a separate paragraph is better. This is what I did in, e.g., [section 2.4.4.1](https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/92.html#value-localizable-string) and [section 2.6.2](https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/92.html#manifest-lang-dir-local). I agree with @aphillips that keeping such examples on separate lines is more readable

Okay, I'll take care of that tomorrow. It is a bit disconcerting that the inline experience can be so variable.

gkellogg:
> > > But, if you think what we have in our doc looks correct, then I won't worry about it.
> > 
> > 
> > I ran in the same problem in the [PR](https://github.com/w3c/pub-manifest/pull/92) on the publication manifest. I think that, for readability's sake, putting the problematic sentences into a separate paragraph is better. This is what I did in, e.g., [section 2.4.4.1](https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/92.html#value-localizable-string) and [section 2.6.2](https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/92.html#manifest-lang-dir-local). I agree with @aphillips that keeping such examples on separate lines is more readable
> 
> Okay, I'll take care of that tomorrow. It is a bit disconcerting that the inline experience can be so variable.

@iherman I instead put it as a statements table, which shows the directionality properly within table cells; this is consistent with how we treat other examples, see what you think.

index.html
```diff
@@ -11781,7 +12008,12 @@ <h2>Value Patterns</h2>
...
-      <li>The values of <code>@value</code>, <code>@language</code> and <code>@type</code> MAY additionally be
+      <li>The values of
+        <code>@value</code>,
+        <code>@language</code>,
+        <code class="changed">@direction</code> and
+        <code>@type</code>
+        MAY additionally be
         an empty <a>map</a> (<a data-cite="JSON-LD11-FRAMING#dfn-wildcard">wildcard</a>),
         an <a>array</a> containing only an empty <a>map</a>,
         an empty <a>array</a> (<a data-cite="JSON-LD11-FRAMING#dfn-match-none">match none</a>)
```
```suggestion
        an empty <a>array</a> (<a data-cite="JSON-LD11-FRAMING#dfn-match-none">match none</a>),
```

index.html
```diff
@@ -12376,6 +12608,14 @@ <h2>Relationship to RDF</h2>
       and <code>false</code>. The JSON-LD 1.1 Processing Algorithms and API specification [[JSON-LD11-API]]
       defines the <a data-cite="JSON-LD11-API#data-round-tripping">conversion rules</a>
       between JSON's native data types and RDF's counterparts to allow round-tripping.</li>
+    <li class="changed">As an extension to the <a data-cite="RDF11-CONCEPTS#data-model">RDF data model</a>,
+      <em>typed literals</em>, where the datatype is `xsd:string`, or <a>language-tagged strings</a>
+      MAY include a <a>base direction</a>.
+      As there is currently no standardized mechanism for representing the <a>base direction</a>
+      of <a>RDF literals</a>, the JSON-LD to RDF transformation loses the base direction.
```
dlongley: Given the following sentence, this one is a bit confusing. Perhaps a slight adjustment:

```suggestion
      of <a>RDF literals</a>, the JSON-LD to standard RDF transformation loses the base direction.
```

index.html
```diff
@@ -12376,6 +12608,14 @@ <h2>Relationship to RDF</h2>
...
+      The <a data-cite="JSON-LD11-API#deserialize-json-ld-to-rdf-algorithm">Deserialize JSON-LD to RDF Algorithm</a>
```
```suggestion
      However, the <a data-cite="JSON-LD11-API#deserialize-json-ld-to-rdf-algorithm">Deserialize JSON-LD to RDF Algorithm</a>
```

index.html
```diff
@@ -12376,6 +12608,14 @@ <h2>Relationship to RDF</h2>
...
+      also provides a means of representing <a>base direction</a>
```
```suggestion
      provides a means of representing <a>base direction</a>
```

index.html
```diff
@@ -12376,6 +12608,14 @@ <h2>Relationship to RDF</h2>
...
+      using non-standardized mechanisms which will preserve round-tripping through RDF.</li>
```
dlongley: Is the goal to represent `direction` in a non-standard way when this flag is used? I would think that is much better than inventing some kind of translation to an RDF pattern that could map back in a different way. (I understand this is an ongoing discussion, just putting this out there).

Edit: I see below what the current options are. Hmm, well, apply suggestion or ignore it. Given that text I would recommend ignoring it ... but it seems like if we adopt the approach below we'll be stuck with it forever. I'd rather the community not take on that technical debt and instead output non-standard RDF with what we believe is the right way to model this stuff -- and see if it gets uptake or not. That's for another issue/discussion though.

```suggestion
      using mechanisms which will preserve round-tripping through non-standard RDF.</li>
```

index.html
```diff
@@ -5151,6 +5160,214 @@ <h3>Using the Document Base for the Default Vocabulary</h3>
...
+    <!--
+    {
+      "@context": {
+        "title": "http://example.org/title",
+        "publisher": "http://example.org/publisher",
+        ####...####
+        ****"@language": "ar",
+        "@direction": "rtl"****
+      },
+      ****"title": "HTML و CSS: تصميم و إنشاء مواقع الويب"****,
```
pchampin:
```suggestion
      ****"title": "HTML و CSS: تصميم و إنشاء مواقع الويب",
```
(the closing 4-stars are present on the next line)

index.html
```diff
@@ -12376,6 +12608,14 @@ <h2>Relationship to RDF</h2>
       and <code>false</code>. The JSON-LD 1.1 Processing Algorithms and API specification [[JSON-LD11-API]]
       defines the <a data-cite="JSON-LD11-API#data-round-tripping">conversion rules</a>
       between JSON's native data types and RDF's counterparts to allow round-tripping.</li>
+    <li class="changed">As an extension to the <a data-cite="RDF11-CONCEPTS#data-model">RDF data model</a>,
+      <em>typed literals</em>, where the datatype is `xsd:string`, or <a>language-tagged strings</a>
```
pchampin:
```suggestion
      <a>language-tagged strings</a>
```
I do not think that `xsd:string` should be mentionned here. This seems to imply that `{"@value": "hello", "@type": "xsd:string", "@direction": "ltr"}` is allowed, which is not the case.

Furthermore, in a direction-aware jsonld-to-rdf mapping, `{"@value": "hello", "@direction": "ltr"}` would *not* produce an`xsd:string`, but rather something in the line of `"hello"^^i18n:_ltr`.

pchampin: Alternative proposal, may be even more accurate:
```suggestion
      literals without an explicit datatype
```

index.html
```diff
@@ -12592,6 +12832,97 @@ <h3>Serializing/Deserializing RDF</h3>
       </dd>
     </dl>
   </section>
+
+  <section class='informative'><h2>The `i18n` Namespace</h2>
+    <p>The `i18n:` namespace is used for describing combinations of <a>language tag</a> and <a>base direction</a> in <a>RDF literals</a>.
+      It is used as an alternative mechanism for describing the [[BCP47]] <a>language tag</a> and <a>base direction</a>
+      of <a>RDF literals</a> that would otherwise be use `xsd:string` or `rdf:langString` datatypes.</p>
```
pchampin:
```suggestion
      of <a>RDF literals</a> that would otherwise use the `xsd:string` or `rdf:langString` datatypes.</p>
```

index.html
```diff
@@ -12592,6 +12832,97 @@ <h3>Serializing/Deserializing RDF</h3>
...
+    <pre class="turtle-dt nohighlight"
+         data-content-type="text/turtle"
+         data-transform="updateExample"
+         data-options="rdfDirection=datatype"
+         data-to-rdf>
+    <!--
+    @prefix ex: <http://example.org/> .
+    @prefix i18n: <https://www.w3.org/i18n#> .
+
+    # Note that this version preserves the base direction using a datatype.
```
pchampin:
```suggestion
    # Note that this version preserves the base direction using a non-standard datatype.
```

index.html
```diff
@@ -12592,6 +12832,97 @@ <h3>Serializing/Deserializing RDF</h3>
...
+      ex:title ****"HTML و CSS: تصميم و إنشاء مواقع الويب"^^i18n:ar_rtl****;
+      ex:publisher ****"مكتبة"^^i18n:ar_rtl****
+    ] .
+    -->
+    </pre>
+    <p>See <a href="#base-direction" class="sectionRef"></a> for more details
+      on using <a>base direction</a> for strings.</p>
+  </section>
+
+  <section class='informative'><h2>The `rdf:CompoundLiteral` class and the `rdf:language` and `rdf:direction` Datatypes</h2>
```
pchampin:
```suggestion
  <section class='informative'><h2>The `rdf:CompoundLiteral` class and the `rdf:language` and `rdf:direction` properties</h2>
```

index.html
```diff
@@ -12592,6 +12832,97 @@ <h3>Serializing/Deserializing RDF</h3>
...
+    <p>This specification defines the `rdf:CompoundLiteral` datatype, which is in the domain
```
pchampin:
```suggestion
    <p>This specification defines the `rdf:CompoundLiteral` class, which is in the domain
```

index.html
```diff
@@ -12592,6 +12832,97 @@ <h3>Serializing/Deserializing RDF</h3>
...
+
+  <section class='informative'><h2>The `i18n` Namespace</h2>
+    <p>The `i18n:` namespace is used for describing combinations of <a>language tag</a> and <a>base direction</a> in <a>RDF literals</a>.
+      It is used as an alternative mechanism for describing the [[BCP47]] <a>language tag</a> and <a>base direction</a>
+      of <a>RDF literals</a> that would otherwise be use `xsd:string` or `rdf:langString` datatypes.</p>
+    <p>Datatypes based on this namespace allow round-tripping of JSON-LD documents using base direction,
+      although the mechanism is not otherwise standardized.</p>
+    <p>The <a data-cite="JSON-LD11-API#deserialize-json-ld-to-rdf-algorithm">Deserialize JSON-LD to RDF Algorithm</a>
+      can be used with the <dfn data-cite="JSON-LD11-API#dom-jsonldoptions-rdfDirection">rdfDirection</dfn>
+      set to `datatype` to generate <a>RDF literals</a> using the `i18n` base to create an IRI
```
pchampin: I would suggest something more specific, such as `i18n-datatype`, in case other implementation come up with alternative datatype-based solution

index.html
```diff
@@ -12592,6 +12832,97 @@ <h3>Serializing/Deserializing RDF</h3>
...
+        The domain of the property is `rdf:CompoundLiteral`.</dd>
+      <dt>`rdf:direction`</dt>
+      <dd>An RDF <a>property</a>.
+        The range of the property is an `rdfs:Literal`, whose value MUST be either `"ltr"` or `"rtl"`.
+        The domain of the property is `rdf:CompoundLiteral`.</dd>
+    </dl>
+
+    <p>The <a data-cite="JSON-LD11-API#deserialize-json-ld-to-rdf-algorithm">Deserialize JSON-LD to RDF Algorithm</a>
+      can be used with the <a>rdfDirection</a>
+      set to `bnode` to generate <a>RDF literals</a> using these properties to
```
pchampin: I would suggest something more specific, such as `compound-literal`, in case other implementation come up with alternative bnode-based solution

index.html
```diff
@@ -12592,6 +12832,97 @@ <h3>Serializing/Deserializing RDF</h3>
...
+    <pre class="turtle-dt nohighlight"
+         data-content-type="text/turtle"
+         data-transform="updateExample"
+         data-options="rdfDirection=bnode"
+         data-to-rdf>
+    <!--
+    @prefix ex: <http://example.org/> .
+
+    # Note that this version preserves the base direction using a bnode structure.
+    # Note that this version preserves the base direction using a bnode structure.
```
pchampin:
```suggestion
```
This line is duplicated.

index.html
```diff
@@ -12592,6 +12832,97 @@ <h3>Serializing/Deserializing RDF</h3>
...
+      of <a>RDF literals</a> that would otherwise be use `xsd:string` or `rdf:langString` datatypes.</p>
+    <p>Datatypes based on this namespace allow round-tripping of JSON-LD documents using base direction,
+      although the mechanism is not otherwise standardized.</p>
+    <p>The <a data-cite="JSON-LD11-API#deserialize-json-ld-to-rdf-algorithm">Deserialize JSON-LD to RDF Algorithm</a>
+      can be used with the <dfn data-cite="JSON-LD11-API#dom-jsonldoptions-rdfDirection">rdfDirection</dfn>
+      set to `datatype` to generate <a>RDF literals</a> using the `i18n` base to create an IRI
+      encoding the <a>base direction</a> along with optional <a>language tag</a> from
+      value objects containing `@direction` by appending to `https://www.w3.org/i18n#`
+      the value of `@language`, if any, followed by an underscore (`"_"`) followed
+      by the value of `@direction`.</p>
```
aphillips: To avoid confusion later, it might be useful to use a language tag with more than one subtag (so the underscore isn't confused with the hyphens used in language tags). In this case, make `ar` -> `ar-EG` (`EG` is the subtag for Egypt), so the serialized form ends up `i18n:ar-EG_rtl`

aphillips:
With the exception of my one comment, the lasted version looks good to me.

index.html
iherman: Minor comment: the URL should be `https://www.w3.org/ns/i18n#`, certainly not a URL at the root of the site.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-10-04-json-ld#section4-1).

- `RESOLVED:  the WG is happy to merge #276, but leaves it to the editors whether they do it now or after the API changes is also in a PR`
- `ACTION:  create issue on values of language maps not including <code>@direction</code> (Gregg Kellogg)`
Benjamin Young: See Issue Syntax#1§1<br/>
Benjamin Young: See Pending PR on <code>@direction</code><br/>
Benjamin Young: no much to discuss; just to make sure everyone is aware of the PR<br/>
Gregg Kellogg: I will only merge it when the tests pass;<br/>
… currently working on my implementation;<br/>
… maybe only merge when we have an API PR.<br/>
… The syntax PR does not address the ability for a language map to contain value objects with <code>@value</code> and <code>@direction.</code><br/>
… Currently they only support strings.<br/>
… So there is no way to override the direction in a language map.<br/>
Ivan Herman: I think we can defer this to a future version.<br/>
Benjamin Young: is there an issue for this already?<br/>
Ivan Herman: raise an issue, so we can make it clear if we decide to postpone it.<br/>
Action #3: create issue on values of language maps not including <code>@direction</code> (Gregg Kellogg)<br/>
Dave Longley: the syntax doc proposes 2 or 3 ways to represent this in RDF in order to roundtrip.<br/>
… We should only recommend one.<br/>
… We would have to maintain them for many years.<br/>
… My preferred option would be to drop the <code>@direction</code> by default,<br/>
… with an option to serialize it in the way we think is the best for future RDF.<br/>
… That would force people to update their RDF lib,<br/>
… but that what we were expecting if a dedicated WG had been created.<br/>
Ivan Herman: half of your wishes are already fulfilled:<br/>
… by default, the <code>@direction</code> is dropped.<br/>
… We picked two options for roundtripping:<br/>
… * the i18n family of datatypes,<br/>
… * the “compound literal” approach (not a very good name).<br/>
… At the meeting in Fukuoka, even around the table,<br/>
… there was no clear consensus about which one was the best option.<br/>
… The idea is to let the community decide which one they want to use.<br/>
Dave Longley: Why not a 3rd option where we add a property to a literal?<br/>
Ivan Herman: because this is not valid RDF.<br/>
… The failure to create a WG shows that the RDF community is not willing to change the core of RDF.<br/>
… The advantage of the two proposed approaches is that they work with the deployed RDF infrastructure,<br/>
… even though they require some application-level knowledge to be used.<br/>
Dave Longley: What I’m hearing is “more technical depts, and no interop”<br/>
… If this is something that people need, and we introduce a hack to get it work,<br/>
… people will adopt the hack.<br/>
Ivan Herman: you can call it a hack, but this is the only thing that people seem ready to accept.<br/>
… Trying to force RDF to change is a bigger hack in my opinion.<br/>
Dave Longley: JSON-LD 1.0 did influence the RDF WG at the time, to extend RDF.<br/>
Ivan Herman: this is the big difference; there was an RDF WG at the time.<br/>
… There is none at the moment, nor can I foresee one in the near future.<br/>
Dave Longley: sometimes, standard follow implementations, rather than the opposite.<br/>
Pierre-Antoine Champin: I’d argue that the proposed solutions are not so hacky from and RDF point of view<br/>
… I prefer the datatype solution<br/>
… but none of them seem a scandal to me<br/>
… if there was an RDF WG working on this, they would consider one of these as the future standard<br/>
Ivan Herman: +1 to pchampin<br/>
Gregg Kellogg: I agree, these solutions are reasonable from the point of view of RDF.<br/>
… Also, the default option is to drop direction.<br/>
… The majority of data will not use direction, and not produce RDF differently than today.<br/>
Dave Longley: dropping it is a good default;<br/>
… but I expect Digital Bazaar being force to keep it in some way.<br/>
… For example, in VC, when the graph is signed, we need to keep the whole information.<br/>
Gregg Kellogg: from a digital signature point of view,<br/>
… I would not consider the direction to be relevant.<br/>
… Would two values differing only by direction be really different?<br/>
Dave Longley: this is scary; that could be turned into a terrible vulnerability.<br/>
Ivan Herman: I would like a resolution to allow Greg to merge the direction PR;<br/>
… this has lingered for too long.<br/>
Proposed resolution: the WG is happy to merge #276, but leaves it to the editors whether they do it now or after the API changes is also in a PR (Ivan Herman)<br/>
Ivan Herman: +1<br/>
Gregg Kellogg: +1<br/>
Harold Solbrig: +1<br/>
Pierre-Antoine Champin: +1<br/>
Benjamin Young: +1<br/>
Tim Cole: +1<br/>
Dave Longley: +0<br/>
Ruben Taelman: +1<br/>
Resolution #2: the WG is happy to merge #276, but leaves it to the editors whether they do it now or after the API changes is also in a PR<br/>
</details>

index.html
```diff
@@ -11766,6 +11990,9 @@ <h2>Value Objects</h2>
     <p>The value associated with the <code>@language</code> key MUST have the
       <a data-cite="BCP47#section-2.1.1">lexical form</a> described in [[BCP47]], or be <a>null</a>.</p>
 
+    <p>The value associated with the <code>@direction</code> key MUST be
+      one of `ltr` or `rtl`, or be <a>null</a>.</p>
```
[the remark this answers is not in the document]
gkellogg: Yup, backquotes work as expected. Not worth changing current usage, but convenient for new work.

index.html
gkellogg: Update both the language tag and namespace URI.

PR #282, opened 2019-10-08
Weaken requirement to normalize language tags

rubensworks:
As mentioned by @gkellogg in https://github.com/w3c/json-ld-api/pull/164, this weakens the requirement for normalizing language tags to lowercase, and makes it a "MAY", just like the RDF spec does.


gkellogg:
@rubensworks I think this is already handled. Can you close this, and related PR in API if you agree?

rubensworks:
Indeed, I can confirm that the language tag lowercasing requirement is now optional in the syntax document. I've also closed https://github.com/w3c/json-ld-api/pull/164.

Issue #81, opened 2018-10-12
Better styling of changes


gkellogg:
All documents include classes that make it easy to see what's changed from 1.0. This could be styled better.

iherman:
Agree with closing

PR #285, opened 2019-10-19
Add control to highlight/de-highlight changes

gkellogg:
at the end of "How to Read this Document".

This may be too subtle, and we might want to figure out how better to control.

For #81.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Nov 4, 2019, 11:48 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fjson-ld-syntax%2F1278eb2f3b30020ae55623eb64c13ba35c078940%2Findex.html%3FisPreview%3Dtrue)

```
[33m📡 HTTP Error 520:[39m [36mhttps://rawcdn.githack.com/w3c/json-ld-syntax/1278eb2f3b30020ae55623eb64c13ba35c078940/index.html[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/json-ld-syntax%23285.)._
</details>

### DIFF
--- a/common/typographical-conventions.html
+++ b/common/typographical-conventions.html
@@ -33,7 +33,8 @@
     and links to the references section.</dd>
   <dt class="changed">Changes from Recommendation</dt><dd>
     Sections or phrases changed from the previous Recommendation
-    are <span class="changed">highlighted</span>.</dd>
+    may be <span class="changed">highlighted</span> using a control
+    in <a href="#how-to-read-this-document" class="sectionRef"></a>.</dd>
 </dl>
 
 <p class="note">Notes are in light green boxes with a green left border and with a "Note" header in green.
--- a/index.html
+++ b/index.html
@@ -213,7 +213,7 @@
     padding: 0.2em;
   }
   .highlight.changed, .show-changes {
-    background-color: rgb(215, 238, 197);
+    background-color: lime;
   }
   .show-changes.selected:before {
     content: "de-highlight";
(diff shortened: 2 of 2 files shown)


common/typographical-conventions.html
```diff
@@ -33,7 +33,8 @@
     and links to the references section.</dd>
   <dt class="changed">Changes from Recommendation</dt><dd>
     Sections or phrases changed from the previous Recommendation
-    are <span class="changed">highlighted</span>.</dd>
+    may be <span class="changed">highlighted</span> using a control
+    in <a href="#how-to-read-this-document" class="sectionRef"></a>.</dd>
 </dl>
```
iherman: While we are at it... I just realized that the highlight backgrounds uses the same (or almost the same) color as the note backgrounds. This may be misleading. We may want to choose another color...

gkellogg:
I'd like to delegate further UI considerations to @BigBlueHat; the basic mechanism is in place.

common/typographical-conventions.html
gkellogg: Uses "lime" now. This is the same as the diff-changes use.

Issue #288, opened 2019-10-29
Terminology of "absolute IRI" differs from RFC 3986 / 3987?


lo48576:
In JSON-LD spec, "absolute IRI" is actually same as "IRI" in RFC 3987 (defined by `IRI` rule), but not "absolute IRI" in RFC 3987 (defined by `absolute-IRI` rule).

> An absolute IRI is defined in [RFC3987] as containing a scheme along with path and optional query and fragment segments.
>
> --- [Section 3.2 (IRIs) in JSON-LD-SYNTAX](https://w3c.github.io/json-ld-syntax/#iris)

> The absolute form of an IRI containing a scheme along with a path and optional query and fragment segments.
>
> --- [Section 1.4 (Terminology) in JSON-LD-SYNTAX](https://w3c.github.io/json-ld-syntax/#terminology)

> IRI reference: Denotes the common usage of an Internationalized Resource Identifier.  An IRI reference may be absolute or relative.  However, the "IRI" that results from such a reference only includes absolute IRIs; any relative IRI references are resolved to their absolute form.  Note that in [RFC2396] URIs did not include fragment identifiers, but in \[RFC3986] fragment identifiers are part of URIs.
>
> [Section 1.3 (Definitions) in RFC 3987](https://tools.ietf.org/html/rfc3987#section-1.3)

> Some protocol elements allow only the absolute form of a URI without a fragment identifier.  For example, defining a base URI for later use by relative references calls for an absolute-URI syntax rule that does not allow a fragment.
>
>     absolute-URI  = scheme ":" hier-part [ "?" query ]
>
>
> --- [Section 4.3 (Absolute URI) in RFC 3986](https://tools.ietf.org/html/rfc3986#section-4.3)

It seems that RFC 3986 (and also 3987) defines hierarchy of some kinds of URIs / IRIs:

```
URI reference         (absolute or relative. fragment part is allowed)
|-- URI               (absolute. fragement is allowed)
|   `-- absolute URI  (absolute. fragment is NOT allowed)
`-- relative URI      (relative. fragment is allowed)
```

Simply saying "URI" (and "IRI") automatically means it is absolute, and it should be called "IRI reference" to include relative IRI.
"Absolute IRI" (defined by `absolute-IRI` rule) seems to be unable to have a fragment.
(However, RFC also saying "absolute form" of IRI references, so RFC itself is confusing...)

So, I think "absolute IRI" in JSON-LD spec and "absolute IRI" in RFC differs.
[...]

PR #289, opened 2019-10-31
Update use of _IRI_ to obviate _absolute IRI_

gkellogg:
... use _IRI reference_ where appropriate, and replace _relative IRI_ with _relative IRI reference_.

For #288.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/289.html" title="Last updated on Nov 2, 2019, 6:57 PM UTC (19ffb44)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/289/44ecfe8...19ffb44.html" title="Last updated on Nov 2, 2019, 6:57 PM UTC (19ffb44)">Diff</a>

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -67,7 +67,7 @@
     An <a>IRI reference</a> may be absolute or
     <a data-lt="relative IRI reference">relative</a>.
     However, the "IRI" that results from such a reference only includes absolute <a>IRIs</a>;
-    any <a>relative IRI references</a> references are resolved to their absolute form.</dd>
+    any <a>relative IRI references</a> are resolved to their absolute form.</dd>
   <dt><dfn data-cite="RFC3987#section-6.5" class="preserve">relative IRI reference</dfn></dt><dd>
     A relative IRI reference is an <a>IRI reference</a> that is relative to some other <a>IRI</a>,
     typically the <a>base IRI</a> of the document.
--- a/common/terms.html
+++ b/common/terms.html
@@ -81,9 +81,9 @@
 <p>Terms imported from [[[RDF11-CONCEPTS]]] [[RDF11-CONCEPTS]], [[[RDF-SCHEMA]]] [[RDF-SCHEMA]], and [[[LINKED-DATA]]] [[LINKED-DATA]]</p>
 <dl class="termlist" data-sort>
   <dt><dfn data-cite="RDF11-CONCEPTS#dfn-base-iri" class="preserve">base IRI</dfn></dt><dd>
-    The <a>base IRI</a> is an <a>absolute IRI</a> established in the <a>context</a>,
+    The <a>base IRI</a> is an <a>IRI</a> established in the <a>context</a>,
     or is based on the <a>JSON-LD document</a> location.
-    The <a>base IRI</a> is used to turn <a>relative IRIs</a> into <a>absolute IRIs</a>.</dd>
+    The <a>base IRI</a> is used to turn <a>relative IRI references</a> into <a>IRIs</a>.</dd>
   <dt><dfn data-cite="RDF11-CONCEPTS#dfn-blank-node" class="preserve">blank node</dfn></dt><dd>
     A <a>node</a> in a <a>graph</a> that is neither an <a>IRI</a>,
     nor a <a>literal</a>.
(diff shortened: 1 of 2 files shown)


TallTed:
s/`Update use if _IRI_ to obviate`/`Update use of _IRI_ to obviate`/

common/terms.html
```diff
@@ -59,18 +59,17 @@
...
-    <a data-lt="relative IRI">relative</a>.
-    However, the "IRI" that results from such a reference
-    only includes <a>absolute IRIs</a>; any <a>relative IRI</a> references are
-    resolved to their absolute form.</dd>
-  <dt><dfn data-cite="RFC3987#section-6.5" class="preserve">relative IRI</dfn></dt><dd>
-    A relative IRI is an <a>IRI</a> that is relative to some other <a>absolute IRI</a>,
+    An <a>IRI reference</a> may be absolute or
+    <a data-lt="relative IRI reference">relative</a>.
+    However, the "IRI" that results from such a reference only includes absolute <a>IRIs</a>;
+    any <a>relative IRI references</a> references are resolved to their absolute form.</dd>
```
TallTed:
```suggestion
    any <a>relative IRI references</a> are resolved to their absolute form.</dd>
```

index.html
```diff
@@ -583,7 +583,7 @@ <h2>Syntax Tokens and Keywords</h2>
         that processing should continue deeper into a JSON data structure.
         This keyword is described in <a class="sectionRef" href="#data-indexing"></a>.</dd>
       <dt><code>@base</code></dt>
-      <dd>Used to set the <a>base IRI</a> against which to resolve those <a>relative IRIs</a>
+      <dd>Used to set the <a>base IRI</a> against which to resolve those <a>relative IRI references</a>
         interpreted relative to the document.
```
TallTed:
```suggestion
        which are otherwise interpreted relative to the document.
```

BigBlueHat:
@gkellogg there's a couple more "absolute IRI" mentions in the ["base IRI" definition](https://github.com/w3c/json-ld-syntax/pull/289/files#diff-8e05e89f61083b17ccc605f6a2697e39R83-R86), but otherwise this looks good.

common/terms.html
```diff
@@ -59,18 +59,17 @@
...
-    <a data-lt="relative IRI">relative</a>.
-    However, the "IRI" that results from such a reference
-    only includes <a>absolute IRIs</a>; any <a>relative IRI</a> references are
-    resolved to their absolute form.</dd>
-  <dt><dfn data-cite="RFC3987#section-6.5" class="preserve">relative IRI</dfn></dt><dd>
-    A relative IRI is an <a>IRI</a> that is relative to some other <a>absolute IRI</a>,
+    An <a>IRI reference</a> may be absolute or
+    <a data-lt="relative IRI reference">relative</a>.
+    However, the "IRI" that results from such a reference only includes absolute <a>IRIs</a>;
+    any <a>relative IRI references</a> are resolved to their absolute form.</dd>
```
pchampin: Shouldn't this be a little more contextualized? For example "JSON-LD processors will resolve all relative IRI references to their absolute form."

gkellogg: This is basically a copy/paste from the definition in RFC3987, as it's not really our term to define. The `data-cite` will cause all references to just go there. The practice for terms that are "imported" from other specs is to basically just summarize their own definitions.

gkellogg:
> @gkellogg there's a couple more "absolute IRI" mentions in the ["base IRI" definition](https://github.com/w3c/json-ld-syntax/pull/289/files#diff-8e05e89f61083b17ccc605f6a2697e39R83-R86), but otherwise this looks good.

But, that's how the base RFC defines it, so trying to invent our own terminology can be more confusing, which I think was partly @lo48576's point. In context, it should be clear what the meaning is, and the term definitions (both in the spec, and in their original locations) should provide all the clarity that is necessary.

Clearly, RFC3987 was sufficiently confusing, that the 1.0 spec got it wrong, but I don't think it's up to us to change that.

gkellogg:
> I stick to my point that some readers might wrongly interpret `<a>IRI</a>` as meaning "either absolute or relative IRI"... But I have no really good alternative to propose...

But, that's how the base RFC defines it, so trying to invent our own terminology can be more confusing, which I think was partly @lo48576's point. In context, it should be clear what the meaning is, and the term definitions (both in the spec, and in their original locations) should provide all the clarity that is necessary.

Clearly, RFC3987 was sufficiently confusing, that the 1.0 spec got it wrong, but I don't think it's up to us to change that.

pchampin:
> Clearly, RFC3987 was sufficiently confusing, that the 1.0 spec got it wrong, but I don't think it's up to us to change that.

Actually, the rest of the RDF 1.1 WG also got it wrong: https://www.w3.org/TR/rdf11-concepts/#note-iris .

@iherman should we make an erratum? (this is not too bad actually, there are only 6 occurences of "relative" and 4 occurences of "absolute".

iherman:
> @iherman should we make an erratum? (this is not too bad actually, there are only 6 occurences of "relative" and 4 occurences of "absolute".

Yes...

Issue #294, opened 2019-11-11
Consistent use of the 'informative' class


pchampin:
Sorry, I know this is late for raising issues :-/

The use of the 'informative' class in the syntax document is not totally consistent, and in some places, badly ambiguous. Generally speaking, I would expect non-normativeness to propagate downward to subsections., especially because there is no "this section is normative" mention to override non-normative-ness.

* §1 is marked as informative, but §1.4 and §1.7 are not (and, IICC, must be considered normative). I think it would be clearer to move §1.4 and §1.7 in another top-level section (in 1.0, Terminology was §2, after an informative §1).

* §3 is marked as informative, and so are its subsections, except for §3.4.

  * If §3.4 is supposed to be informative (my faboured interpretation), it should be explicitly marked as such, for the sake of consistency with its sibbling sections;
  * otherwise, I think we should *not* mark §3 as informative, and move its introdictory text into its own informative subsection.

* 4 is not marked as informative, although all its subsections are marked as informative, and it contain some introductory text, which IMHO should not be interpreted as normative. Also, all subsubsections (except for 4.1.4, for nor good reason) are marked as informative, which I think is redundant. I suggest adding the 'informative' class to §4, and removing it from §4.*.* 

I will make a PR to to this effect in the evenening (French timezone).

iherman:
I certainly believe we should be consistent and we aren't. 

- I believe the _only_ normative sections are §8, §9, and §10. I do not think §1.4 and §1.7 are; they are part of the introduction and help the reader to navigate into §8 and §9, but the definitions are only normative in those.

- Although you are right that the 'informative' qualification is automatically valid for subsection by default, I do prefer to have that term added to each subsection. It is mostly done but, indeed, not only §4 should be set as informative, but that remark should also be present in §4.1.4. And yes, §3.4 should indeed be marked informative as well.

Whether we adopt this style (ie, marking subsections as informative) is our choice but we should indeed be consistent.

Thanks!

pchampin:
> I do not think §1.4 and §1.7 are; they are part of the introduction and help the reader to navigate into §8 and §9

I tend to agree with you about §1.7. On the other hand §1.4 contains the definitions of JSON-LD specific terms, which are linked to by all mentions of those terms in all documents... And both are *explicitly* marked as normative (`class="normative").

> Although you are right that the 'informative' qualification is automatically valid for subsection by default, I do prefer to have that term added to each subsection.

[...]

PR #295, opened 2019-11-11
Clean informative sections

pchampin:
Addresses #294 .

Two remarks about this PR:

* due to some sections not being normative anymore, 5 references have moved from normative references to non-normative references:

  * INFRA
  *JSON-LD10 (now duplicated with JSON-LD, btw)
  * LINKED DATA
  * RFC5234 (ANBF)
  * WEBIDL

  I don't think any of this is a problem for the SYNTAX document, but others might disagree.

* Should term mentions in the rest of the document (and other documents) point to the normative definition in §9 rather than to §1.4? It seems to make more sense, but on the other hand, definitions in §1.4 are more explicative and readable, and they point to examples in sections 3 and 4... Plus, I don't have the respec skill to do it (I tried and I failed).

  An alternative would be to add, in each definition where this is relevant, a link to the corresponding subsection of §9?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/295.html" title="Last updated on Nov 12, 2019, 11:05 PM UTC (c1878e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/295/411f708...c1878e1.html" title="Last updated on Nov 12, 2019, 11:05 PM UTC (c1878e1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12477,10 +12477,16 @@ the data type to be specified explicitly with each piece of data.</p>
     </dd>
     <dt><code>@language</code></dt><dd>
       The <code>@language</code> keyword MAY be aliased and MAY be used as a key in a <a>value object</a>.
-      Its value MUST be a <a>string</a> with the <a data-cite="BCP47#section-2.1.1">lexical form</a> described in [[BCP47]] or be <a>null</a>..
+      Its value MUST be a <a>string</a> with the <a data-cite="BCP47#section-2.1.1">lexical form</a> described in [[BCP47]] or be <a>null</a>.
       <p>The unaliased <code>@language</code> MAY be used as a key in a <a>context definition</a>,
         or as the value of the <code>@container</code> key within an <a>expanded term definition</a>.</p>
-      <p>See <a class="sectionRef" href="#index-maps"></a> for a further discussion.</p>
+      <p>See <a class="sectionRef" href="#string-internationalization"></a>, <a class="sectionRef" href="#language-maps"></a>.</p>
+    </dd>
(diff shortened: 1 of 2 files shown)


gkellogg:
It's handy to have all definitions in the definitions section. We should probably have references to their normative definitions in §9. Mostly, people will click on it to find out what is meant by the term, and this is what the definition list is intended for. The grammar section is not as easy to read to find intent.

index.html
```diff
@@ -11761,7 +11765,7 @@ <h1>JSON-LD Grammar</h1>
   <section id="terms" class="normative">
     <h2>Terms</h2>
 
-    <p>A <a>term</a> is a short-hand <a>string</a> that expands
+    <p>A <a>term</a> is a short-hand <a>string</a> that MAY expand
```
```suggestion
    <p>A <a>term</a> is a short-hand <a>string</a> that SHOULD expand
```

pchampin: "MAY" was previously used in section 1.4, so I reflected it there as is.

I'm not sure that either MAY or SHOULD are a good match here. Whether a term *can* be expanded does not depends on the implementation, and an implementation MUST expand in some circumstances (e.g. expand algorithm), and MUST NOT expand it in other (e.g. compact algorithm).

I think it is best to revert this line to its original version, or keep "may" but in lowercase?

gkellogg: Okay, we'll revert, then.

gkellogg:
Fixes #294

PR #307, opened 2019-11-21
Update example to actually require `@graph`.

### LINKED ISSUE #300: Named Graphs: confusing "embedding doesn't work as each node object references the other"
There is a confusing line in the 4.9 Named Graphs section: https://github.com/w3c/json-ld-syntax/blob/master/index.html#L9157 It says: 
> In this case, embedding doesn't work as each node object references the other. 

But above, in the [example 94](https://www.w3.org/TR/json-ld11/#example-94-referencing-an-unidentified-node) the same case is given with embedding working just fine with blank node id. 

This is confusing, because i have an impression that by replacing blank node identifier with a regular iri ("http://manu.sporny.org/about#manu" ) this example shows that embedding works here just fine.

If I am missing something, perhaps it worth a rewording?

gkellogg:
Fixes #300.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/307.html" title="Last updated on Nov 21, 2019, 9:51 PM UTC (adbde64)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/307/db875c2...adbde64.html" title="Last updated on Nov 21, 2019, 9:51 PM UTC (adbde64)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -9088,13 +9088,11 @@ the data type to be specified explicitly with each piece of data.</p>
         {
           "@id": "http://manu.sporny.org/about#manu",
           "@type": "Person",
-          "name": "Manu Sporny",
-          "knows": "https://greggkellogg.net/foaf#me"
+          "name": "Manu Sporny"
         }, {
           "@id": "https://greggkellogg.net/foaf#me",
           "@type": "Person",
-          "name": "Gregg Kellogg",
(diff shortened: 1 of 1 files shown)


pchampin:
I guess we can close this one, as you have merged mine...

Issue #309, opened 2019-11-22
Alias statements about `@base` are inconsistent


hsolbrig:
Section 4.6.1 says "Each of the JSON-LD keywords, except for `@context`, may be aliased to application-specific keywords."  Section 9.16 Keywords says, however, that "The `@base` keyword MUST NOT be aliased, and MAY be used as a key in a context definition."  Can both of these statements be true?

gkellogg:
It seems more clarification is needed. Within a context, keyword aliases may not be used. Within the body of a document, all keywords other than `@context` May be aliased. Of course, `@base` only appears in the context, along with some other keywords.

hsolbrig:
How about something along the lines of "Any JSON-LD keyword that can appear in a _node object_ may be aliased with the exception of `@context`"?

gkellogg:
It’s not just node objects, but really anywhere outside of the context. I thought we had wording about this, maybe obscured from 1.0.

gkellogg:
See #308; we should simplify these and normatively prohibit using aliased keywords in contexts or term definitions.

azaroth42:
Resolved by #312

Issue #308, opened 2019-11-22
Confusing statement on aliasing keywords


hsolbrig:
The sentence "Other than for @type, properties of expanded term definitions where the term is a keyword result in an error. " (Section 4.1.6) is difficult to understand.  Would it be possible to add an example of an "@type" and another keyword to clarify the intent?

gkellogg:
There's a note in Aliasing Keywords: "Aliased keywords MUST NOT be used within a context, itself.", but it's not normative, and shouldn't use normative language.

There are a number of places in keyword definitions where we say --term-- MUST NOT be aliased, and we probably shouldn't make this statement. It's not that there's something special about a keyword (other than `@context`) that prevents it from being aliased, it's that nothing in a context definition can be aliased. The grammar and term definition sections should make this clear normatively. We should remove statements we've added about not aliasing keywords other than `@context` and clarify that keywords may be aliased in the body of a JSON-LD document exclusive of a context or term definition.

gkellogg:
There is already a reference to `#using-set-with-type` which has such an example. I've updated the text to also say that there are examples there, in addition to further detail.

azaroth42:
Editorial work resolved by #312

PR #312, opened 2019-11-24
Keyword aliases

gkellogg:
* Add clarification to keyword aliasing and remove MUST NOT be aliased description, as this implies that doing so would lead to an exception, which it does not.
* Remove normative language from a note in the Aliasing Keywords Section.
* Be explicit about where keywords can and cannot be aliased in the normative Keywords section.
* Reference `#using-set-with-type` for examples on using `@type` in a context.

For #308 and for #309.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/312.html" title="Last updated on Nov 25, 2019, 9:50 PM UTC (d230aee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/312/4ef8738...d230aee.html" title="Last updated on Nov 25, 2019, 9:50 PM UTC (d230aee)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12421,10 +12421,10 @@ the data type to be specified explicitly with each piece of data.</p>
     <a>list objects</a>,
     <a>set objects</a>, and
     <a>nested properties</a>
-    all <a>keywords</a> other than `@context` MAY be aliased.
-    The `@context` MUST NOT bee aliased.
+    <a>keyword</a> aliases MAY be used instead of the corresponding <a>keyword</a>, except for `@context`.
+    The `@context` <a>keyword</a> MUST NOT bee aliased.
     Within <a>local contexts</a> and <a>expanded term definitions</a>,
-    <a>keywords</a> MAY NOT be aliased, other than to create a <a>keyword</a> alias.</p>
+    <a>keyword</a> aliases MAY NOT used.</p>
 
   <dl data-sort>
     <dt><code>@base</code></dt><dd>
--- a/index.html
+++ b/index.html
@@ -3101,7 +3101,8 @@
     result in an error.
     <span class="changed">Unless the <a>processing mode</a> is set to `json-ld-1.0`,
       there is also an exception for <code>@type</code>;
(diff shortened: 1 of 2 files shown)


gkellogg:
The keywords section made claims about the ability to alias or not keywords, which isn't entirely true. The only keyword that MUST NOT be aliased is `@context`. Others MAY be aliased, but might not be appropriate outside of a context or term definition, but it's not an error to create such an alias. Keywords which MAY be aliased (such as `@id`) still can't be aliased within term definitions.

index.html
```diff
@@ -12410,13 +12414,25 @@ <h2>Keywords</h2>
...
     this section describes where each <a>keyword</a> may appear within different JSON-LD structures.</p>
 
+  <p>Within
+    <a>node objects</a>,
+    <a>value objects</a>,
+    <a>graph objects</a>,
+    <a>list objects</a>,
+    <a>set objects</a>, and
+    <a>nested properties</a>
+    all <a>keywords</a> other than `@context` MAY be aliased.
```
pchampin:
```suggestion
    <a>keyword</a> aliases MAY be used instead of the corresponding <a>keyword</a>, except for `@context`.
```

index.html
```diff
@@ -12410,13 +12414,25 @@ <h2>Keywords</h2>
...
+    The `@context` MUST NOT bee aliased.
```
pchampin:
```suggestion
    The `@context` <a>keyword</a> MUST NOT bee aliased.
```

index.html
```diff
@@ -12410,13 +12414,25 @@ <h2>Keywords</h2>
...
+    Within <a>local contexts</a> and <a>expanded term definitions</a>,
+    <a>keywords</a> MAY NOT be aliased, other than to create a <a>keyword</a> alias.</p>
```
pchampin:
```suggestion
    <a>keyword</a> aliases MAY NOT used.</p>
```
I think the "other than to create a kw alias" adds confusion...

PR #350, opened 2020-05-06
Added a note on superseding 1.0 into the SOTD

iherman:
At the instruction of PlH I have added a note into the SOTD about this. To make things quicker and avoid problems, I have changed the version of the document already on /TR.

(Oops, I realized I made a spelling mistake in the name of the branch:-)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/350.html" title="Last updated on May 6, 2020, 7:29 PM UTC (0a27e4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/350/6c4be3d...0a27e4c.html" title="Last updated on May 6, 2020, 7:29 PM UTC (0a27e4c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -318,6 +318,8 @@
     <a href="https://json-ld.org/playground/">live JSON-LD playground</a> that is capable
     of demonstrating the features described in this document.</p>
 
+  <p>This specification is intended to <a href='https://www.w3.org/2019/Process-20190301/#rec-rescind'>supersede</a> [[[JSON-LD10]]] [[JSON-LD10]]. </p>
+
   <section>
     <h2>Set of Documents</h2>
     <p>This document is one of three JSON-LD 1.1 Recommendations produced by the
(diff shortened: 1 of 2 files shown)


gkellogg:
@iherman I presume we need the same changes for API and Framing.

iherman:
For framing it is not relevant, we did not have an earlier rec.

You are right about the API, I forgot. I will do it

gkellogg:
And, it sounds like you've updated the to-be-published document(s) in-place, so no need to update the publication snapshot?

iherman:
> And, it sounds like you've updated the to-be-published document(s) in-place, so no need to update the publication snapshot?

Only if we want to be a purist:_) I think we can do without.

Do not merge it yet, b.t.w., I do it in the API right now and I may change the sentence sligthly. I will add a ping here.

iherman:
Ok, done now. And the API is also done.

PR #360, opened 2020-07-08
Finalize dates on the errata page

iherman:
This should be merged when the final document is published


iherman:
Did it by hand, not realizing that this PR is here...

PR #415, opened 2023-06-23
Add structured suffix

OR13:
***
<a href="https://pr-preview.s3.amazonaws.com/OR13/json-ld-syntax/pull/415.html" title="Last updated on Jul 17, 2023, 6:10 PM UTC (790ff22)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/415/54f6d1d...OR13:790ff22.html" title="Last updated on Jul 17, 2023, 6:10 PM UTC (790ff22)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -13540,7 +13540,7 @@ the data type to be specified explicitly with each piece of data.</p>
         </dl>
       </dd>
       <dt>Person &amp; email address to contact for further information:</dt>
-      <dd>Ivan Herman &lt;ivan@w3.org&gt;</dd>
+      <dd>Pierre-Antoine Champin &lt;pierre-antoine@w3.org&gt;</dd>
       <dt>Intended usage:</dt>
       <dd>Common</dd>
       <dt>Restrictions on usage:</dt>
@@ -13692,9 +13692,7 @@ the data type to be specified explicitly with each piece of data.</p>
       information.
       -->
       <dt>Contact</dt>
-      <dd>
-        Ivan Herman &lt;ivan@w3.org&gt;
-      </dd>
+      <dd>Pierre-Antoine Champin &lt;pierre-antoine@w3.org&gt;</dd>
 
       <!-- 
       Person (including contact information) authorized to change this
--- a/index.html
+++ b/index.html
@@ -13738,7 +13738,7 @@ the data type to be specified explicitly with each piece of data.</p>
       </dd>
       
     </dl>
-    <section id="iana-examples" class="informative">
+    <section id="structured-extension-examples" class="informative">
       <h3>Examples</h3>
       <p>The following examples illustrate different ways in which this structured suffix may be used
         to request and respond with more specific content types for a resource.
--- a/index.html
+++ b/index.html
@@ -13614,7 +13614,7 @@ the data type to be specified explicitly with each piece of data.</p>
         are enclosed in double quotes (<code>"</code>).</p>
     </section>
   </section>
-  <section>
+  <section id="structured-extension-ld-json">
     <h3>+ld+json</h3>
 
     <p>This section will be submitted to the Internet Engineering Steering
@@ -13672,7 +13672,7 @@ the data type to be specified explicitly with each piece of data.</p>
       <dt>Fragment identifier considerations</dt>
       <dd>
         <p>
-          Fragment identifiers used with <a href="#ld-json">+ld+json</a>
+          Fragment identifiers used with <a href="#structured-extension-ld-json">+ld+json</a>
           are treated as in RDF syntaxes, as per
           <a data-cite="RDF11-CONCEPTS#section-fragID">RDF 1.1 Concepts and Abstract Syntax</a>
           [[RDF11-CONCEPTS]].
@@ -13872,8 +13872,6 @@ the data type to be specified explicitly with each piece of data.</p>
     <li>The <a>processing mode</a> is now implicitly `json-ld-1.1`, unless set
       explicitly to `json-ld-1.0`.</li>
     <li>Improve notation using <a>IRI</a>, <a>IRI reference</a>, and <a>relative IRI reference</a>.</li>
-    <!--li>Allow further structured subtypes of `application/ld+json` by using
-      `+ld+json` as a suffix for a new base type.</li-->
     <li>Warn about forward-compatibility issues for terms of the form (`"@"1*ALPHA`).</li>
     <li>When creating an `i18n` datatype or `rdf:CompoundLiteral`, <a>language tags</a> are
       normalized to lower case to improve interoperability between implementations.</li>
@@ -13911,6 +13909,8 @@ the data type to be specified explicitly with each piece of data.</p>
     <li>Misclaneous spelling and grammar fixes.</li>
     <li>Regenerated definition list, which does not attempt to filter out unused definitions.</li>
     <li>Fixed typo in <a href="#alternate-document-location" class="sectionRef"></a> where "HTTP document" was intended to be "HTML document".</li>
+    <li>Allow further structured extensions of `application/ld+json` by using
+      `+ld+json` as a structured media extension. See <a href="#structured-extension-ld-json"></a>.</li>
   </ul>
 </section>
 
--- a/index.html
+++ b/index.html
@@ -13424,193 +13424,346 @@ the data type to be specified explicitly with each piece of data.</p>
 
   <p>This section has been submitted to the Internet Engineering Steering
     Group (IESG) for review, approval, and registration with IANA.</p>
+  
+  <section>
+    <h3>application/ld+json</h3>
+    <dl>
+      <dt>Type name:</dt>
+      <dd>application</dd>
+      <dt>Subtype name:</dt>
+      <dd>ld+json</dd>
+      <dt>Required parameters:</dt>
+      <dd>N/A</dd>
+      <dt>Optional parameters:</dt>
+      <dd>
+        <dl>
+          <dt><code>profile</code></dt>
+          <dd>
+            <p>A non-empty list of space-separated URIs identifying specific
+              constraints or conventions that apply to a JSON-LD document according to [[RFC6906]].
+              A profile does not change the semantics of the resource representation
+              when processed without profile knowledge, so that clients both with
+              and without knowledge of a profiled resource can safely use the same
+              representation. The <code>profile</code> parameter MAY be used by
+              clients to express their preferences in the content negotiation process.
+              If the profile parameter is given, a server SHOULD return a document that
+              honors the profiles in the list which it recognizes,
+              and MUST ignore the profiles in the list which it does not recognize.
+              It is RECOMMENDED that profile URIs are dereferenceable and provide
+              useful documentation at that URI. For more information and background
+              please refer to [[RFC6906]].</p>
+            <p>This specification defines six values for the <code>profile</code> parameter.</p>
+            <dl>
+              <dt><code>http://www.w3.org/ns/json-ld#expanded</code></dt>
+              <dd>To request or specify <a data-lt="expanded document form">expanded JSON-LD document form</a>.</dd>
+              <dt><code>http://www.w3.org/ns/json-ld#compacted</code></dt>
+              <dd>To request or specify <a data-lt="compacted document form">compacted JSON-LD document form</a>.</dd>
+              <dt><code>http://www.w3.org/ns/json-ld#context</code></dt>
+              <dd>To request or specify a <a data-lt="context document">JSON-LD context document</a>.</dd>
+              <dt><code>http://www.w3.org/ns/json-ld#flattened</code></dt>
+              <dd>To request or specify <a data-lt="flattened document form">flattened JSON-LD document form</a>.</dd>
+              <dt><code>http://www.w3.org/ns/json-ld#frame</code></dt>
+              <dd>To request or specify a <a data-lt="frame">JSON-LD frame document</a>.</dd>
+              <dt><code>http://www.w3.org/ns/json-ld#framed</code></dt>
+              <dd>To request or specify <a data-lt="framed document form">framed JSON-LD document form</a>.</dd>
+            </dl>
+            <p>All other URIs starting with <code>http://www.w3.org/ns/json-ld</code>
+              are reserved for future use by JSON-LD specifications.</p>
+            <!--p>Other specifications MAY create further structured subtypes
+              by using `+ld+json` as a suffix for a new base subtype, as in
+              `application/example+ld+json`.
+              Unless defined otherwise, such subtypes use the same
+              fragment identifier behavior as `application/ld+json`.</p-->
+            <p>Other specifications may publish additional `profile` parameter
+              URIs with their own defined semantics.
+              This includes the ability to associate a file extension with a `profile` parameter.</p>
+            <p>
+              When used as a <a data-cite="RFC4288#section-4.3">media type parameter</a> [[RFC4288]]
+              in an <a data-cite="rfc7231#rfc.section.5.3.2">HTTP Accept header</a> [[RFC7231]],
+              the value of the <code>profile</code> parameter MUST be enclosed in quotes (<code>"</code>) if it contains
+              special characters such as whitespace, which is required when multiple profile URIs are combined.</p>
+            <p>When processing the "profile" media type parameter, it is important to
+              note that its value contains one or more URIs and not IRIs. In some cases
+              it might therefore be necessary to convert between IRIs and URIs as specified in
+              <a data-cite="RFC3986#section-5.1">section 3 Relationship between IRIs and URIs</a>
+              of [[RFC3987]].</p>
+          </dd>
+        </dl>
+      </dd>
+      <dt>Encoding considerations:</dt>
+      <dd>See <a data-cite="RFC8259#section-11">RFC&nbsp;8259, section 11</a>.</dd>
+      <dt id="iana-security">Security considerations:</dt>
+      <dd>See <a data-cite="RFC8259#section-12">RFC&nbsp;8259, section 12</a> [[RFC8259]]
+        <p>Since JSON-LD is intended to be a pure data exchange format for
+          directed graphs, the serialization SHOULD NOT be passed through a
+          code execution mechanism such as JavaScript's <code>eval()</code>
+          function to be parsed. An (invalid) document may contain code that,
+          when executed, could lead to unexpected side effects compromising
+          the security of a system.</p>
+        <p>When processing JSON-LD documents, links to remote contexts and frames are
+          typically followed automatically, resulting in the transfer of files
+          without the explicit request of the user for each one. If remote
+          contexts are served by third parties, it may allow them to gather
+          usage patterns or similar information leading to privacy concerns.
+          Specific implementations, such as the API defined in the
+          JSON-LD 1.1 Processing Algorithms and API specification [[JSON-LD11-API]],
+          may provide fine-grained mechanisms to control this behavior.</p>
+        <p>JSON-LD contexts that are loaded from the Web over non-secure connections,
+          such as HTTP, run the risk of being altered by an attacker such that
+          they may modify the JSON-LD <a>active context</a> in a way that
+          could compromise security. It is advised that any application that
+          depends on a remote context for mission critical purposes vet and
+          cache the remote context before allowing the system to use it.</p>
+        <p>Given that JSON-LD allows the substitution of long IRIs with short terms,
+          JSON-LD documents may expand considerably when processed and, in the worst case,
+          the resulting data might consume all of the recipient's resources. Applications
+          should treat any data with due skepticism.</p>
+        <p>As JSON-LD places no limits on the IRI schemes that may be used,
+          and vocabulary-relative IRIs use string concatenation rather than
+          IRI resolution, it is possible to construct IRIs that may be
+          used maliciously, if dereferenced.</p>
+      </dd>
+      <dt>Interoperability considerations:</dt>
+      <dd>Not Applicable</dd>
+      <dt>Published specification:</dt>
+      <dd>http://www.w3.org/TR/json-ld</dd>
+      <dt>Applications that use this media type:</dt>
+      <dd>Any programming environment that requires the exchange of
+        directed graphs. Implementations of JSON-LD have been created for
+        JavaScript, Python, Ruby, PHP, and C++.
+      </dd>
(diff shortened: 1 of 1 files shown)


OR13:
I have joined the group and signed the IPR agreement, but this change is indeed substantive.

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+          <dt>Magic number(s):</dt>
+          <dd>Not Applicable</dd>
+          <dt>File extension(s):</dt>
+          <dd>.jsonld</dd>
+          <dt>Macintosh file type code(s):</dt>
+          <dd>TEXT</dd>
+        </dl>
+      </dd>
+      <dt>Person &amp; email address to contact for further information:</dt>
+      <dd>Ivan Herman &lt;ivan@w3.org&gt;</dd>
```
gkellogg, line 13547: Contact should probably be @pchampin, who is the current WG Team Contact. But, I wonder if W3C has a policy for this, as team members also come and go over time.

gkellogg (approved):

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
 
   <p>This section has been submitted to the Internet Engineering Steering
     Group (IESG) for review, approval, and registration with IANA.</p>
+  
+  <section>
+    <h3>application/ld+json</h3>
```
TallTed, line 13429:
```suggestion
    <h3>Media Type Registration: `application/ld+json`</h3>
```

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+      </pre>
+      <p>Requests the server to return the requested resource as JSON-LD
+        in both <a>compacted document form</a>
+        and <a>flattened document form</a>.
+        Note that as whitespace is used to separate the two URIs, they
+        are enclosed in double quotes (<code>"</code>).</p>
+    </section>
+  </section>
+  <section>
+    <h3>+ld+json</h3>
```
TallTed, line 13618:
```suggestion
    <h3>Structured Suffix Registration: `+ld+json`</h3>
```

TallTed (changes requested):

gkellogg:
We tried to do this in JSON-LD 1.0, but IANA wasn't prepared to allow such structured suffix registrations at the time. @msporny was the primary person driving this, IIRC.

msporny:
> We tried to do this in JSON-LD 1.0, but IANA wasn't prepared to allow such structured suffix registrations at the time. @msporny was the primary person driving this, IIRC.

Yes, and that led to this draft that allows this sort of structured suffix registration:

https://www.ietf.org/archive/id/draft-ietf-mediaman-suffixes-04.html

The specification above is reaching IETF Last Call, and so we need to test to see if the new spec will allow what we tried to register many years ago.

"+ld+json" is the first dry run of this new ability to register media types and structured suffixes with multiple plus signs.

dlongley (approved):

peacekeeper (approved):
Yes this is useful.

gkellogg:
@pchampin can we clear the IPR exception?

gkellogg:
@OR13, you may need to set up your W3C account to know about your GitHub ID for the IPR validation to work automatically. I see you're listed as a participant, so I'm going ahead and marking it as complete.

gkellogg:
(Never mind, it re-evaluated on it's own).

OR13:
I don't intend to accept the change suggestions that add extra text to h2/h3 and make the sidebar look bad.

TallTed:
> I don't intend to accept the change suggestions that add extra text to h2/h3 and make the sidebar look bad.

I guess I will have to wait until PR#415 is merged (assuming it's accepted), and submit my own subsequent PR for the heading changes, to see what the WG prefers and the Editors will therefore apply.

gkellogg:
The PR is on the agenda for the CH call this Wednesday. When considering @TallTed’s suggestion, I’d consider other similar registrations and look for some form of conformity.

gkellogg:
The format of the [YAML Media Type](https://www.ietf.org/archive/id/draft-ietf-httpapi-yaml-mediatypes-09.html) seems to match @TallTed’s suggested change.

OR13:
The format of the existing registration matches mine.

OR13:
The yaml spec isn't using respec, so it in general looks a lot nicer...

gkellogg (changes requested):
From [the June 28th meeting](https://json-ld.org/minutes/2023-06-28/#28):

>  I had a look at w3c/json-ld-syntax#415, but I'm a bit uncomfortable with the duplication with the application/ld+json registration.

This should probably be re-worked for less duplication, and needs an addition to [Changes since the Recommendation of 16 July 2020](https://w3c.github.io/json-ld-syntax/#changes-from-rec1).

OR13:
@gkellogg are you able to suggest changes on the PR?

gkellogg:
I can probably push commits to your branch, or do a PR into your branch.

OR13:
The GitHub UI also has a suggest changes feature, which would allow me to commit your suggestion directly to the branch

gkellogg:
> The GitHub UI also has a suggest changes feature, which would allow me to commit your suggestion directly to the branch

Yes, IIUC, it only allows changes to lines that were already in the branch; this could be used for some changes, but not to add anything to the "Changes since ..." section.

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
 
   <p>This section has been submitted to the Internet Engineering Steering
     Group (IESG) for review, approval, and registration with IANA.</p>
+  
+  <section>
```
gkellogg, line 13428: Allow predictable references to this section.

```suggestion
  <section id="media-type-ld-json">
```

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
-    <a data-cite="RDF11-CONCEPTS#section-fragID">RDF 1.1 Concepts and Abstract Syntax</a>
-    [[RDF11-CONCEPTS]].</p>
+    <p>This registration is an update to the original definition
+      for <a data-cite="JSON-LD10#application-ld-json">application/ld+json</a>
+      in [[JSON-LD10]].</p>
 
-  <p>This registration is an update to the original definition
-    for <a data-cite="JSON-LD10#application-ld-json">application/ld+json</a>
-    in [[JSON-LD10]].</p>
+    <section id="iana-examples" class="informative">
```
gkellogg, line 13567: Make section identifier unique, and add anchor for original identifier.

```suggestion
    <section id="media-type-examples" class="informative"><span id="iana-examples"></span>
```

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+      -->
+      </pre>
+      <p>Requests the server to return the requested resource as JSON-LD
+        in both <a>compacted document form</a>
+        and <a>flattened document form</a>.
+        Note that as whitespace is used to separate the two URIs, they
+        are enclosed in double quotes (<code>"</code>).</p>
+    </section>
+  </section>
+  <section>
```
gkellogg, line 13617: Predictable section identifier.

```suggestion
  <section id="structured-extension-ld-json">
```

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+      Person (including contact information) authorized to change this
+      suffix registration.
+      -->
+      <dt>Author/Change controller</dt>
+      <dd>
+        W3C
+      </dd>
+      
+    </dl>
+    <section id="iana-examples" class="informative">
```
gkellogg, line 13741:
```suggesetion
    <section id="structured-extension-examples" class="informative">
```

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+      </dd>
+
+      <!-- 
+      Generic processing of fragment identifiers for any type employing
+      this syntax should be described here.
+      -->
+      <dt>Fragment identifier considerations</dt>
+      <dd>
+        <p>
+          Fragment identifiers used with <a href="#ld-json">+ld+json</a>
```
gkellogg, line 13675:
```suggestion
          Fragment identifiers used with <a href="#structured-extension-ld-json">+ld+json</a>
```

index.html
```diff
@@ -13758,6 +13909,8 @@ <h2>Changes since Recommendation of 16 July 2020</h2>
     <li>Misclaneous spelling and grammar fixes.</li>
     <li>Regenerated definition list, which does not attempt to filter out unused definitions.</li>
     <li>Fixed typo in <a href="#alternate-document-location" class="sectionRef"></a> where "HTTP document" was intended to be "HTML document".</li>
+    <li>Allow further structured extensions of `application/ld+json` by using
+      `+ld+json` as a structured media extension. See <a href="#structured-extension-ld-json"></a>.</li>
```
gkellogg, line 13913: @OR13 I pushed this directly to your branch, which I couldn't do through the UI. Feel free to suggest updates to it. Note the changes to the section identifier which are in separate update suggestions.

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+      <dt>Type name:</dt>
+      <dd>application</dd>
+      <dt>Subtype name:</dt>
+      <dd>ld+json</dd>
+      <dt>Required parameters:</dt>
+      <dd>N/A</dd>
+      <dt>Optional parameters:</dt>
+      <dd>
+        <dl>
+          <dt><code>profile</code></dt>
```
gkellogg, line 13440: My suggestion to reduce duplication for the `profile` parameter is to create a separate sub-section for that parameter, and change this to a reference with "See description in <a href="#profile-parameter"></a>", or something similar. That's a bit too much to attempt through the web UI, but I can push a commit to the source branch if this is agreeable.

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+              it might therefore be necessary to convert between IRIs and URIs as specified in
+              <a data-cite="RFC3986#section-5.1">section 3 Relationship between IRIs and URIs</a>
+              of [[RFC3987]].</p>
+          </dd>
+        </dl>
+      </dd>
+      <dt>Encoding considerations:</dt>
+      <dd>See <a data-cite="RFC8259#section-11">RFC&nbsp;8259, section 11</a>.</dd>
+      <dt id="iana-security">Security considerations:</dt>
+      <dd>See <a data-cite="RFC8259#section-12">RFC&nbsp;8259, section 12</a> [[RFC8259]]
```
gkellogg, line 13496: The preferred practice is to have a "Security Considerations" section and reference this from the media type registration. We do have such a section, but it references this. We should basically move this content to the existing "Security Considerations" section and replace this with a reference back using "See <a href="#security"></a>".

index.html
gkellogg: @pchampin, either you should be the contact, or someone more permanent at W3C.

```suggestion
      <dd>Pierre-Antoine Champin &lt;pierre-antoine@w3.org&gt;</dd>
```

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+      </dd>
+
+      <!-- 
+      Person (including contact information) to contact for further
+      information.
+      -->
+      <dt>Contact</dt>
+      <dd>
+        Ivan Herman &lt;ivan@w3.org&gt;
+      </dd>
```
gkellogg, line 13729: @pchampin, either you should be the contact, or someone more permanent at W3C.

```suggestion
      <dd>Pierre-Antoine Champin &lt;pierre-antoine@w3.org&gt;</dd>
```

index.html
```diff
@@ -13424,193 +13424,346 @@ <h2>IANA Considerations</h2>
...
+
+      <!-- 
+      Security considerations shared by media types employing this
+      structured syntax must be specified here.  The same requirements
+      for media type security considerations given in Section 4.6 apply
+      here, with the exception that the option of not assessing the
+      security considerations is not available for suffix registrations.
+      -->
+      <dt>Security considerations</dt>
```
gkellogg, line 13689: See comment above on referencing security considerations and moving content there.

gkellogg:
@OR13 I made some inline suggestions, and comments about further changes that should probably be made directly to your branch. If you agree, I'll do that. It would still leave some duplication, which could be managed by further extracting into common sections, but improves it quite a bit, otherwise.

Note that I expect we'll follow this PR up with some more editorial changes.

index.html
OR13: I will merge with confirmation from @pchampin

index.html
OR13: thats fine, I am not attached to any of the specific language here, only to completing the process.... so that it can be taken as a proper dependency of the VC WG and DID WG

index.html
OR13: I don't have an opinion on this, related to JSON-LD's use of the media type parameter.

There are other structures suffixes (like +cose) that inherit media type parameters from their normal media type...

cose-type for application/cose and +cose... 

profile for application/ld+json and +ld+json ?

not my call to make... but I agree there should be clarity provided on this topic.

OR13: go ahead and take over the branch completely, I just came to start the party, you all will probably need to finish it.

index.html
OR13: Same comment as above, you can push whatever you like to the branch.

OR13:
@gkellogg thanks for your suggestions, I merged the ones I could, and I authorize you to push whatever you like to the branch, and or reword any amount of this in subsequent PRs, I am mostly hoping to have a stable reference ahead of IETF 117 in case this comes up, so if its possible to merge (even with notes , issue markers or warnings before then, I appreciate it). 

In the case it is not merged by 117 I will update IETF, that this is still in progress and might need to be updated depending on how WGLC goes for multiple suffixes draft.

gkellogg:
This should just need @pchampin's approval, and decision on the suggestions for updating the contact in the registrations.

OR13:
Awesome!

index.html
pchampin: I think it makes sense to put my contact details here.

pchampin (approved):

gkellogg (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-28-vcwg-special#section1-4) on 2023-11-28

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 1.4. Request profile parameter from `application/vc` (issue vc-data-model#1363)

_See github issue [vc-data-model#1363](https://github.com/w3c/vc-data-model/issues/1363)._




**Brent Zundel:** The current path we are on relies on media types with multiple suffixes.  
… The spec for multiple suffixes in IETF has been more contentious than anticipated.  
… And possibly a whole lot more work than anticipated.  
… This issue is another path to sidestep this.  

**Orie Steele:** This issue can be used with or without multiple suffixes.  
… Our current registrations require multiple suffixes, even though this isn't possible at present.  
… Even for application/vc the subtype can contain media type parameters.  
… Those parameters can express characteristics of the media type.  
… Specifying the text type is a common usage.  
… This tracks the intended use of the profile parameter.  
… We should provide guidance on the use of profile.  
… We could say not to use it.  
… We could say to use it in the same way as ld+json.  

**Brent Zundel:** Thank you for clarifying the relationship between the parameters.  
… We can register media types with multiple suffixes.  
… The current interpretation of a+b+c is that a is the prefix and b+c is the suffix.  

> *Orie Steele:* we can request registrations that rely on multiple suffixes, but they cannot be registered, until their interpretation is clear... it is a mistake to request registrations that we know cannot be approved.

**Ted Thibodeau Jr.:** The first thing after the slash isn't what matters.  
… The subtype is what's after the last + sign.  
… There are multiple layers of subtyping.  
… I don't know why the IETF did this.  

> *Orie Steele:* if you have questions regarding media types, you can ask the media type list. see [https://datatracker.ietf.org/doc/draft-ietf-mediaman-suffixes/](https://datatracker.ietf.org/doc/draft-ietf-mediaman-suffixes/).

**Ted Thibodeau Jr.:** You process as many of the subtypes that you understand.  
… This is complex and esoteric.  

> *Orie Steele:* +1 TallTed, its complicated : (.

**Ted Thibodeau Jr.:** The problem with multiple + signs is that the interpretation is undefined.  
… There may be many more drafts.  
… We can specify what our media types mean but those will only apply to people who understand our spec.  
… It's to let people who don't understand what's to the left of the +s to still work with what's to the right of the +s.  

> *Orie Steele:* here is the repo: [https://github.com/ietf-wg-mediaman/suffixes](https://github.com/ietf-wg-mediaman/suffixes) ... TallTed, eager to see your PR : ).

**Manu Sporny:** I am the editor of that spec.  
… There is only one remaining issue.  
… It's about whether to register all the things in between.  

> *Dave Longley:* for "application/vc+ld+json" processing is: "application", then "json", then "ld+json", then "vc+ld+json" ... another way to understand this subclass processing is: "application/square+rectangle+shape" ... you can understand any "application" content, then, if you want, more specifically any "shape", ... then any "rectangle" ... "square".

**Manu Sporny:** The chair of the WG wants to take it to last call.  
… I don't think the multiple structured suffixes draft is in a place that it's not going to be ratified soon.  
… Ted is right that we can just specify what our media types mean.  
… We don't need to change anything about the media types used in the spec.  
… The profile problem is problemantic.  
… Many people don't prrocess it.  

> *Orie Steele:* My recommendation would be to not depend on multiple suffixes, we raised a PR to remove the dependency from vc-jose-cose ... [https://github.com/w3c/vc-jose-cose/pull/186](https://github.com/w3c/vc-jose-cose/pull/186).

> *Orie Steele:* related registration for [https://www.iana.org/assignments/media-types/application/ld+json](https://www.iana.org/assignments/media-types/application/ld+json).

> *Orie Steele:* related specification Manu mentioned: [https://www.w3.org/TR/activitystreams-core/#media-type](https://www.w3.org/TR/activitystreams-core/#media-type).

**Manu Sporny:** You can have multiple different IETF registrations stomp over it.  
… I realize we used it in JSON-LD. I wish we hadn't.  
… We can tell people not to use it.  
… What we've learned is that people don't implement it correctly.  
… It's often destroyed or mangled as it goes through HTTP servers.  
… We can only pick one filename suffix.  
… We can't use profile to do that.  

> *Orie Steele:* feels like implementing profile correctly, is easier than using multiple suffixes consistently.

> *Dmitri Zagidulin:* `@orie` - agree.

> *Ted Thibodeau Jr.:* *sighs* people who don't conform to specs (e.g., by misconfiguring servers or otherwise breaking the `profile` option) should not be justification for not conforming to specs!

> *Manu Sporny:* It's not easier to implement profile correctly, Orie. :) -- We have multiple decades of experience now that people don't implement it correctly.

> *Dave Longley:* as long as multiple suffixes follows the rule that each subtype fits within the same wider syntax, things are kept clean (and correct) ... i think trouble has come from people thinking it's for enveloping, when it isn't.

> *Manu Sporny:* ^ yes, that.

> *Manu Sporny:* (to what dlongley said).

**Dmitri Zagidulin:** I want to agree with Orie that hierarchical multiple suffixes is way more trouble than it's worth.  
… I don't think there's any need for hierarchical multiple suffixes.  

> *Orie Steele:* +1 ivan.

> *Dmitri Zagidulin:* +1 to at risk.

**Ivan Herman:** If we want to keep multiple suffixes, we should put the whole concept as being at risk in the document.  
… Put it as at risk.  
… We can see what happens at the IETF.  
… If it goes through at IETF, we can remove the at risk.  
… Otherwise we can reconsider.  
… We have to defend ourselves.  

**Brent Zundel:** We're relying on multiple suffixes but not normatively pointing to them.  

> *Manu Sporny:* Note that JSON-LD has gone through multiple RECs w/ this in there: [https://w3c.github.io/json-ld-syntax/#structured-extension-ld-json](https://w3c.github.io/json-ld-syntax/#structured-extension-ld-json).

**Ivan Herman:** On the other hand, we are supposed to register media types we use during CR and we can't do it now.  
… We can't ignore this.  
… That's why we have to mark it as at-risk.  

**Orie Steele:** I spoke to our W3C liaison at the IETF - Martin.  
… It might be fine to go into CR that way.  
… To become a recommendation, we have to finish the IETF work.  
… It's OK to go into CR with the references are they are.  
… If we have to correct this, it would be good to be able to do this without going through CR.  
… If marking it as at-risk helps, let's do that.  
… I don't know how the DID spec got to Rec with multiple suffixes.  
… If this isn't resolved before REC, I will formally object.  

> *Orie Steele:* thats not true manu.

> *Orie Steele:* I added the +ld+json structured suffix.

**Manu Sporny:** The DID and JSON-LD specs have gone through REC with multiple structured suffixes.  
… You can treat it as an opaque thing.  
… We can change it to application/vc if we need to.  
… But that's doing the wrong thing.  
… We should fix the IETF's broken suffixes.  
… We should stop routing around this damage at the IETF.  

_See github pull request [json-ld-syntax#415](https://github.com/w3c/json-ld-syntax/pull/415)._




> *Brent Zundel:* [https://www.w3.org/TR/did-core/#application-did-ld-json](https://www.w3.org/TR/did-core/#application-did-ld-json).

> *Orie Steele:* seems like the W3C is responsible for the broken behavior, and I agree that it needs to be fixed.

> *Orie Steele:* I am trying to fix it FWIW.

**Brent Zundel:** The DID spec said that it will be registered as soon as the IETF spec is done.  

**Ivan Herman:** The only thing I don't want is to go there and not even mention the problem.  
… If we include a note like we did in the DID spec, that may be enough.  

**Brent Zundel:** What are the steps forward for the profile parameter issue?  
… We could prohibit their use or say how to use them.  

**Manu Sporny:** We tell people how to use profile parameters correctly.  
… Expect people to break this and it not to work correctly across the ecosystem.  
… We should warn them that it has never worked out well when people rely on the profile parameter.  

> *Ivan Herman:* +1 to manu's direction.

**Brent Zundel:** We don't need to include doomsaying.  
… We would need someone willing to be assigned to the issue.  
… If you feel strongly that doing nothing is the right approach, then please volunteer to be assigned.  
… If noone is assigned, then by default, we will be doing nothing about this.  

> *Orie Steele:* the media types impact the "verification algorithm".

**Ivan Herman:** All these options we are talking about are editorial?  

> *Orie Steele:* are they still editorial?

**Brent Zundel:** The media types affect the verification algorithm.  
… We don't have anyone willing to be assigned.  

> *Orie Steele:* the reason I am not volunteering is that we are removing the dependency from vc-jose-cose.

</details>

PR #418, opened 2023-08-09
Update references to RFC3987 with RDF12-CONCEPTS.

### LINKED ISSUE #355: Kill "IRI"?
The heavy use of "IRI" as a term, referencing RFC 3987, raises a number of thorny issues and actually makes JSON-LD out-of-sync with the latest developments at W3C. See https://www.w3.org/International/wiki/IRIStatus.

If the spec continues to use the term, then there should probably be a section added to the spec to explain how JSON-LD proposes to solve the IRI problems such as inconsistent use of punycode and percent encoding. I suspect it would be simpler to use "URL" everywhere, per W3C recommendations, with a simple (foot)note explaining that the intent of "URL" is to encompass internationalization as originally envisioned by the IRI effort, but will track the work of the new URL working group for the particulars.

gkellogg:
Updates other RDF references to RDF 1.2 versions.
Depends on an update to RDF12-CONCEPTS to export the "IRI reference" term. Temporarily skips the check for consistence of common files.

Fixes #355.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/418.html" title="Last updated on Aug 9, 2023, 8:34 PM UTC (c6f43d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/418/cee83dc...c6f43d6.html" title="Last updated on Aug 9, 2023, 8:34 PM UTC (c6f43d6)">Diff</a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,11 +20,12 @@ jobs:
     - name: Install dependencies
       run: bundle install
 
-    - name: "Verify common files are consistent"
-      run: |
-        git remote add -f b https://github.com/w3c/json-ld-wg.git
-        git remote update
-        git diff --exit-code remotes/b/main -- common
+    # Remove temorarily
+    #- name: "Verify common files are consistent"
+    #  run: |
+    #    git remote add -f b https://github.com/w3c/json-ld-wg.git
+    #    git remote update
+    #    git diff --exit-code remotes/b/main -- common
 
     - name: Verify examples are consistent
       run: bundle exec rake test
--- a/common/terms.html
+++ b/common/terms.html
@@ -66,34 +66,13 @@
   </dd>
 </dl>
 
-<p>Terms imported from [[[RFC3987]]] [[RFC3987]]</p>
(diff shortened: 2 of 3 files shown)


gkellogg:
This is a test of using the RDF Concepts definition of IRI instead of referencing RFC3987.

dlongley (approved):
Seems ok to me -- other than that RDF12-CONCEPTS is still a working draft and these are normative references. I can't recall how we ought to proceed with these sorts of interim changes, so I'm deferring to those more knowledgeable.

gkellogg:
The idea of using RDF12-CONCEPTS was to start doing the updates that would naturally come when updating to JSON-LD 1.2, which would come after RDF 1.2. Some of the concepts we need to clarify, such as character space and text direction, are better defined in RDF 1.2, so it makes sense to start updating those references now.

Generally, references can lag, so that as an ED, we can reference WDs. By the time JSON-LD starts going rec-track again, RDF 1.2 should be in CR. Certainly, other necessary updates will be based on changes in RDF 1.2.

Blakgirl:
Gkelloge

pchampin:
I'm catching up on long overdue PR reviews...
It seems to me that, given the current plan to amend the current REC with class1-3 changes, this PR should not be merged just yet, right?

pchampin:
I propose to close this PR, as it now conflicts with other PRs affecting common files (and should be rendered moot by https://github.com/w3c/json-ld-wg/issues/170

